PR #47, opened 2016-02-08
pull request for xpdAcq issues

chiahaoliu:
move plotting out from saving function

Need to refactor _feature_gen function later

### DIFF
--- a/xpdacq/analysis.py
+++ b/xpdacq/analysis.py
@@ -36,6 +36,7 @@ get_events = _get_obj('get_events')
 get_images = _get_obj('get_images')
 
 
+
 def bt_uid():
     return bt.get(0).md['bt_uid']
 
@@ -81,14 +82,13 @@ def _feature_gen(header):
     exp_time = _timestampstr(header.start.time)
     return '_'.join([exp_time, f_name])
 
-def _timestampstr(timestamp):
-    time = str(datetime.datetime.fromtimestamp(timestamp))
-    date = time[:10]
-    hour = time[11:16]
-    m_hour = hour.replace(':','-')
-    timestampstring = '_'.join([date,hour])
-    #corrected_timestampstring = timestampstring.replace(':','-')
-    return timestampstring
+def _timestampstr(timestamp, hour=False):
+    ''' convert timestamp to strftime formate '''
+    if not hour:
+        timestring = datetime.date.fromtimestamp(float(timestamp)).strftime('%Y_%m_%d')
+    elif hour:
+        timestring = datetime.date.fromtimestamp(float(timestamp)).strftime('%Y_%m_%d-%H')
+    return timestring
 
 def save_last_tiff():
     save_tif(db[-1])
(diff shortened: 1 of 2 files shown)


xpdacq/analysis.py
```diff
@@ -89,16 +90,14 @@ def _timestampstr(timestamp):
     #corrected_timestampstring = timestampstring.replace(':','-')
     return timestampstring
 
-def save_last_tif():
+def save_last_tiff():
```
tacaswell: In general, if you are going to do rename like this you should probably include

```python
save_last_tif = save_last_tiff
```

to make sure you do not break any code on the beamline computers.

chiahaoliu: good lesson, @tacaswell

tacaswell:
You can keep pushing commits to this branch to update your PR.

sbillinge:
@tacaswell I am still trying to understand our workflow.

If I understand correctly, MASTER in xpdacq/xpdacq now contains deployed code, and nothing should be pushed there until it has actually been tested in the production environment (in our case XPD).  Does this mean that we need a branch in xpdacq/xpdacq against we make pull requests, maybe we could call it ``ready-to-test`` or something.  Then we can test ``ready-to-test`` at XPD during a maintenance period.  If it works, then we can merge it to master. Then the workflow could be like this:

```
1) read-to-test is branched from MASTER on xpdacq/xpdacq
2) all developers fork their own copy of xpdacq/xpdacq which comes with two branches, MASTER and ready-to-test
3) we all clone myname/xpdacq
4) we make a local branch or branches for developing (do we branch off MASTER or RTT?)
5) we do lots of local testing for syntax errors and local stuff.  Ideally we have test cases and we make them pass, but we haven't learned how to put that into our workflow yet.  It would be good to start that for the V1.0 release milestone, but not before.
6) we periodically push our repo to our fork and request comments and feedback from others who we have linked to the fork.
7) When we are ready we make a pull request to the ``read-to-test`` branch of xpdacq
8) this is tested, modified then merged.

It seems like the git workflow is a bit easier than this (no ready-to-test branch) but then is there a way to test the version of the code in the pull request in the production environment before we merge it?

sbillinge:
@tacaswell btw, I sent a couple of invitations to Dan to join this but he hasn't responded.  Does he not want to be looped in (fine with me if you are our point of contact in DAMA)?  Otherwise it would be good to get his feedback too.

tacaswell:
What in the 'ready-to-test' branch?

tacaswell:
Sigh, I should finish reading before commenting no?

tacaswell:
It is a matter of naming, but most projects invert the names, 'master' is where everything gets merged into and every so often you tag a release and create a maintenance branch which only gets bug fixes.  In this case it matters far more what is _deployed_ to the production environment (which really should be a tagged version) than the state of any given branch.

This is also thing that automated testing should really be able to help with.

PR #91, opened 2016-02-22
just wrapped user input statement

sbillinge:
to make it easier for testing.

I should have done this on a branch but I effed up!


sbillinge:
tested at XPD so pulling into xpdAcq/master

PR #92, opened 2016-02-23
fist attmpt to clean B_DIR and other objects

chiahaoliu:
This pull request focuses on:
  - cleaning ambiguity on naming `home_dir` and `b_dir`
  - central config to B_DIR

`_start_beamtime` and `_end_beamtime` are now working correctly but many other places need to be refactored later.

Note: `999-load.py` in `xpdConfig` and `xpd_profile_file` are the same. Just an extra copy.

### DIFF
--- a/xpdConfig/999-load.py
+++ b/xpdConfig/999-load.py
@@ -135,7 +135,7 @@ class DataPath(object):
 hostname = socket.gethostname()
 if hostname == 'xf28id1-ws2':
     # real experiment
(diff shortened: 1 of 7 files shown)


sbillinge:
After discussion (and more cleaning) I would like to consider replacing this pull-request with #93 which used this pull-request as a starting point, but refactored a bit.  I will be pushing more commits to that pull request shortly too.

chiahaoliu:
:+1:  I will close this pull request for now.

PR #93, opened 2016-02-24
Chiahaoliu work beamtime config

sbillinge:
Simon's efforts at handling globals.  It is passing unit tests now so everything is importing ok, but much work still needs to be done.

I left the bluesky imports as they were as I can't test them, but I feel good about having all the globals in one file and one class, and now DataPath is only in one place.  

Still need to work on making it work with the 999-load for xpd that Tim wrote.


chiahaoliu:
Pull in first for rebasing branch. Test on XPD will be done as soon as possible once XPD computer gets back.

PR #95, opened 2016-02-25
work on unittesf of endbeamtime

chiahaoliu:
This pull request is just a way to share my unittest. Please feel free to comment.

finished test delete_home_dir_tree(). Fix a potentially dangerous logic


sbillinge:
@chiahaoliu I talked to Ben about desirable export workflows.  I think there are many that the users may want.  For now let's rename our export function to _bundle_and_archive_userarea() and put it into beamtimeSetup.  Just make it write the bundled export file to the archive area. 

If we want another function that does a similar bundling for the user we can write that later.  This way no big tar file gets written to the user-area.

Can you build this into the functional_tests.py.  Could you also please rename that to beamtimesetup_functional_tests.py too?

Thanks,

chiahaoliu:
Sure, I will do it

chiahaoliu:
close. this is outdated.

PR #98, opened 2016-02-26
0226 dark subtraction

chiahaoliu:
first trial of dark subtraction

Codes turn to be super long for this functionality as a few exceptions should be captured. Logic could possibly be better for a robust code.

unittest to this functionality will be actively updated in this pull request.

PR #101, opened 2016-03-01
Work glbl config

chiahaoliu:
Because of rebase, commits belonging to pull request #98 are shown here at this time. Production commits are those after commit `fd01e7a`.

### DIFF
--- a/tests/test_beamtimeSetup.py
+++ b/tests/test_beamtimeSetup.py
@@ -3,13 +3,14 @@ import os
 import shutil
 import yaml
 from time import strftime
-from xpdacq.xpdacq import _areaDET
-from xpdacq.xpdacq import _tempController
-from xpdacq.xpdacq import _shutter
-from xpdacq.xpdacq import _bdir
-from xpdacq.xpdacq import _cdir
-from xpdacq.xpdacq import _hdir
-from xpdacq.xpdacq import _hostname
+from xpdacq.glbl import _areaDET
+from xpdacq.glbl import _tempController
+from xpdacq.glbl import _shutter
+# these are now included in glbl class
+#from xpdacq.xpdacq import _bdir
+#from xpdacq.xpdacq import _cdir
+#from xpdacq.xpdacq import _hdir
+#from xpdacq.xpdacq import _hostname
 from xpdacq.glbl import glbl
 
 import xpdacq.beamtimeSetup as bts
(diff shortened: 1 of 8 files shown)


sbillinge:
I checked this out but it doesn't seem to be passing its tests (errors in test_beamtimeSetup)

tacaswell:
```python
if __name__ == '__main__':
```

Is a python language feature https://docs.python.org/3/library/__main__.html

sbillinge:
I get that.  But by using this trick in the ipython context, does it make the module namespace share the namespace of the main ipython session....is that the point?

sbillinge:
@chiahaoliu this still seems to be failing test_beamtimeSetup tests with import errors.

chiahaoliu:
May I know the error? I was awared that there are known issues coming from ipython profile, but I forgot to discuss with you before you left.

sbillinge:
cannot import name AREA_DET

On Tue, Mar 1, 2016 at 10:28 PM, Timothy Liu <notifications@github.com>
wrote:

> May I know the error? I was awared that there are known issues coming from
> ipython profile, but I forgot to discuss with you before you left.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/xpdAcq/xpdAcq/pull/101#issuecomment-191037338>.
>



-- 
Prof. Simon Billinge
Applied Physics & Applied Mathematics
Columbia University
500 West 120th Street
Room 200 Mudd, MC 4701
New York, NY 10027
Tel: (212)-854-2918 (o) 851-7428 (lab)

Condensed Matter Physics and Materials Science Dept.
Brookhaven National Laboratory
P.O. Box 5000
Upton, NY 11973-5000
(631)-344-5661

email: sb2896 at columbia dot edu
home: http:// <http://nirt.pa.msu.edu/>bgsite.apam.columbia.edu/

chiahaoliu:
Yes, that is the error I should have mentioned. The main idea of this `global` trick is that `AREA_DET` and other objects are created when profile are loaded. Using `py.test` might not be able to capture this ipython profile behavior properly. I can try to make objects in `setup` section, but this might again mean we are not testing the production code.

tacaswell:
A good practice is to default values to `None`.  That way the names exist and you can check if they are set via `is None`.

chiahaoliu:
do you mean to set default value of object to `None` and selectively import objects depending on values?

tacaswell:
No, just check them for being not-`None` before using them.

chiahaoliu:
I think I might misunderstand some parts of `global` method. I will ask you tomorrow if you will be in @tacaswell

chiahaoliu:
replace with #105. Most updated one

PR #110, opened 2016-03-03
0302_fix known error found at xpd

chiahaoliu:
fix some missing imports we found today at XPD and move `_open_shutter()` to correct order inside `prun`

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -20,11 +20,12 @@ import time
 import datetime
 import numpy as np
 import copy
+import sys
 
 from bluesky.plans import Count
 from bluesky import Msg
-
 from bluesky.plans import AbsScanPlan
+
 from xpdacq.utils import _graceful_exit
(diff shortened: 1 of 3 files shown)


xpdacq/xpdacq.py
```diff
@@ -138,12 +139,12 @@ def prun(sample,scan,**kwargs):
     scan - scan metadata object
     **kwargs - dictionary that will be passed through to the run-engine metadata
     '''
-    if scan.shutter: _open_shutter()
     scan.md.update({'xp_isprun':True})
     light_cnt_time = scan.md['sc_params']['exposure']
     expire_time = glbl.dk_window
     dark_field_uid = validate_dark(light_cnt_time, expire_time)
     if not dark_field_uid: dark_field_uid = dark(sample, scan, **kwargs)
+    if scan.shutter: _open_shutter()
```
tacaswell: it is bad practice to do these one-line `if: stmt` lines.

PR #123, opened 2016-03-09
Version with bluesky objects mocked in simulation mode

sbillinge:
still more cleaning needed on globals, but this is loading properly and running even when in xpdacq is called, so now we can test prun and other functions with unittests.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -149,7 +149,7 @@ def _load_calibration(calibration_file_name = None):
     config_dict : dict
     a dictionary containing calibration parameters calculated from SrXplanar
     '''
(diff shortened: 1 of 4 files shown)


xpdacq/xpdacq.py
```diff
@@ -198,7 +198,7 @@ def prun(sample,scanplan,**kwargs):
     if not dark_field_uid:
         dark_field_uid = dark(sample, scanplan, **kwargs)
```
chiahaoliu: Based on my test at XPD, directly apply the sam scanplan object might lead to some problems. One possible situation is we will run a "dark" Tramp, if the there is no dark frame with the same exposure time as light frame.

PR #126, opened 2016-03-10
Fix yamify dark logic

chiahaoliu:
fix yamify dark

not finished yet, still need to fix unittest

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -107,6 +107,8 @@ def _unpack_and_run(sample,scan,**kwargs):
 
 def _read_dark_yaml():
     dark_yaml_name = glbl.dk_yaml
+    if not os.path.isfile(dark_yaml_name):
+        return None # doesn't exit. Return None and trigger all the way to collect the first dark
     with open(dark_yaml_name, 'r') as f:
         dark_scan_list = yaml.load(f)
     return dark_scan_list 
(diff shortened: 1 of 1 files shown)


xpdacq/xpdacq.py
```diff
@@ -127,8 +129,10 @@ def validate_dark(light_cnt_time, expire_time, dark_scan_list = None):
         dark_field_uid : str
             uid to qualified dark frame
     '''
-    if not dark_scan_list: dark_scan_list= _read_dark_yaml()
-    if len(dark_scan_list) > 0:
+    if not dark_scan_list:
+        dark_scan_list= _read_dark_yaml()
+    #if len(dark_scan_list) > 0:
```
sbillinge: why are you changing this line?  It is testing something different...it is testing that the list exists but is empty, not if it doesn't exist.

I am not sure that I really like your fix of this bug.  The bug was really that something went wrong during the initialization sequence of icollection, so I would rather that the bug-fix fixed the initialization.  In other words, something happens to create the dark yaml file during start beamtime sequence.    I would definitely feel much more comfortable with that.

Having _read_yaml crash when there is no yaml is actually better behavior in a way because it lets us know that something is wrong with the intialization (as it did here) that we need to fix.  Maybe we find other problems with the initialization logic when we do that.

chiahaoliu:
I see,so I guess the behavior we want to is to add initiating dark_yaml logic during `_start_beamtime` and give a warning if program couldn't find dark yaml.

But should we stop user from using any functionality in `xpdAcq` if we found that he didn't follow proper workflow?

chiahaoliu:
close this PR and replace with another

PR #127, opened 2016-03-10
tests for auto-dark-collection now working

sbillinge:
using mock objects.

Finally got the tests to pass in the mocked prun case.


chiahaoliu:
The unittest doesn't pass under an environment without `xpdSim`, which might be the our future plan.

Here is the error I got:
```
    from bluesky.run_engine import RunEngine
E   ImportError: No module named 'bluesky'
```
Maybe we can put real `bluesky` dependency like imports and code block:
```
xpdRE = RunEngine()
xpdRE['group'] = GROUP
...
``` 
under `if hostname == BEAMLINE_HOST_NAME` statement inside `glbl.py`?

PR #132, opened 2016-03-11
Add load user yaml

chiahaoliu:
finished load user yaml. Unittest passes as well.


chiahaoliu:
I delete outdated branch a while ago and create a similar one, but git seems to confuse history wand include history. Conflicts have been carefully handled.

PR #135, opened 2016-03-13
finished cleaning dependencies on collection side

chiahaoliu:
make logic of differentiating experiment/simulation happen during startup process


sbillinge:
@chiahaoliu I'm having trouble resolving conflicts here and getting tests to pass.  I think I can fix it but not right now.  Please don't work on this and I will push when I have it working.

chiahaoliu:
I also found I didn't successfully add and push commits I did earlier today. Last commit is to share the last few changes. I will wait for your push :)

chiahaoliu:
Updated logic based on discussion through phone this afternoon has been pushed. Code passes all unittests and running simulation(run `ipython --profie=collectiopn` locally with new `999-load.py`) is also possible.

xpdacq/glbl.py
```diff
@@ -79,39 +61,92 @@ class glbl():
...
-    xpdRE.md['owner'] = OWNER
-    xpdRE.md['beamline_id'] = BEAMLINE_ID
-    xpdRE.md['group'] = GROUP
-    
-    if hostname != BEAMLINE_HOST_NAME:
-        shutter = mock_shutter()
+    '''
+    # logic to assign correct objects depends on simulation or real experiment
+    hostname = socket.gethostname()
+    if hostname == BEAMLINE_HOST_NAME:
```
tacaswell: This is a _super_ sketchy/brittle way to do this.  A better way might to look for an env like `'XPD_REAL_COLLECTION_ENV'` environmental variable that you can make sure you set as part of a bash script you use to launch ipython

sbillinge: @tacaswell Yes, I agree.  For now this works, but it would be good to do it with an environment variable.

chiahaoliu: We need support for advanced stuffs . When could the next meeting :p ?

Issue #129, opened 2016-03-10
change where Exposure class is created


sbillinge:
This horrible logic of setting up scan.md items then running RE then unsetting them is a strong indication that we have the wrong design, and brings back bad memories of when we were trying to do this with gs.RE (do you remember that! :)

Actually what is wrong is that these medata items are not ScanPlan attributes, they are Scan attributes. or perhaps Exposure attributes.  If we set them at the right level, we should never have to unset them later.  We create an Exposure instance inside _get_light_images.  Perhaps we either need a Scan object in prun or to create the Exposre instance in prun and pass it down.

chiahaoliu:
:+1:

chiahaoliu:
after meeting, we both agree on calling this `Exposure` class mentioned here as `Scan` class, since we don't really have access to information at exposure level ( mostly are machine readback values)

`Scan` class will be created in `beamtime.py`. It inherits metadata from`ScanPlan` and `Sample`. One thing definitely wanted inside is like `Scan.md = Union(ScanPlan.md, Sample.md)`. Detailed

PR #138, opened 2016-03-16
Work exposure class

chiahaoliu:
Finished implementing new `Scan` object inside `prun`, as mentioned in issue #129.
Code pass all unittests and is able to run under simulation environment at this stage. 

I am inheriting the logic from PR #135 in this PR but I can edit accordingly after we decide how we are going to handle and load global objects. This edit should be fairly quick.

### DIFF
--- a/tests/test_validate_dark.py
+++ b/tests/test_validate_dark.py
@@ -13,8 +13,8 @@ from xpdacq.beamtimeSetup import _start_beamtime, _end_beamtime
 from xpdacq.xpdacq import validate_dark, _yamify_dark, prun, _read_dark_yaml
 #from xpdacq.mock_objects import Cam
 
-shutter = glbl.shutter
-shutter.put(1)
+#shutter = glbl.shutter
+#shutter.put(1)
 #glbl.area_det.number_of_sets.put = MagicMock(return_value=1)
 #glbl.area_det.cam = Cam()
 #glbl.area_det.cam.acquire_time.put = MagicMock(return_value=1)
--- a/xpd_profile_file/999-load.py
+++ b/xpd_profile_file/999-load.py
@@ -19,14 +19,90 @@
 import os
 import socket
 from xpdacq.glbl import glbl
+from unittest.mock import MagicMock
+from xpdacq.mock_objects import mock_shutter, mock_livetable
+#from xpdacq.glbl import _areaDET
+#from xpdacq.glbl import _tempController
+# test at xpd
+#from xpdacq.glbl import _shutter
+#from xpdacq.glbl import _verify_write
+#from xpdacq.glbl import _LiveTable
+#from xpdacq.glbl import _dataBroker
+#from xpdacq.glbl import _getEvents
+#from xpdacq.glbl import _getImages
 from xpdacq.beamtimeSetup import _start_beamtime, _end_beamtime
 from xpdacq.beamtime import *
 
+#_areaDET(pe1c)
+#_tempController(cs700)
+
+#######
+# watch out, some testing here
+#######
+#_shutter(shctl1)
+
+# try to keep attribute names the same as the way we are goona use. Can be
+# changed later
+
+
+# logic to make correct objects depends on simulation or real experiment
+hostname = socket.gethostname()
+#if hostname == BEAMLINE_HOST_NAME:
+if hostname == glbl.beamline_host_name:
+    # real experiment
+    simulation = False
+    from bluesky.run_engine import RunEngine
+    from bluesky.register_mds import register_mds
+    from bluesky import Msg
+    from bluesky.plans import Count
+    from bluesky.plans import AbsScanPlan
+
+    xpdRE = RunEngine()
+    xpdRE.md['owner'] = glbl.owner
+    xpdRE.md['beamline_id'] = glbl.beamline_id
+    xpdRE.md['group'] = glbl.group
+    register_mds(xpdRE)
+else:
+    simulation = True
+    BASE_DIR = os.getcwd()
+    ARCHIVE_BASE_DIR = os.path.join(BASE_DIR,'userSimulationArchive')
+    xpdRE = MagicMock()
+    # magic mock objects
+    Msg = MagicMock()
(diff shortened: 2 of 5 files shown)


sbillinge:
mmmm, not sure how I feel about this.  I don't think that we should have mocking in the production code.  Have to think about this some more.  I also worry that mocking objects that haven't been properly assigned will lead to confusing and unexpected behavior for users.

chiahaoliu:
I agree that mocking is not the best behavior in production code and I was very hesitant too. This mocking is only needed for run unittest on `test_validate_dark.py` since we call `prun` inside this unittest but `prun` needs objects like `xpdRE` or so to run. In either beamline or local computer, we shouldn't need it because we will run `icollection` or `ipyhton --profile=collection` either way (and test on this behavior passed)

sbillinge:
it almost certainly means we have our design wrong.  I am still puzzling
what is happening with the imports and the globals.  I think thinks are
pretty messed up.  Perhaps we can talk about it a bit, but I think that we
probably can clean nearly everything out of ``999-load`` except the logic
that checks if this is a new start or a restart, then put it all in
``glbl`` with statements that decide on the behavior of the ``glbl`` class
attributes depending on whether it is run from XPD or otherwise, with
genuine objects in the former and mock objects in the latter, then in the
modules where they are used we load them from the ``glbl`` class and not by
importing them.

Will that work?  If it does, it is much cleaning and we clean everything up
once and for all.

S

On Tue, Mar 15, 2016 at 10:05 PM, Timothy Liu <notifications@github.com>
wrote:

> I agree that mocking is not the best behavior in production code and I was
> very hesitant too. This mocking is only needed for run unittest on
> test_validate_dark.py since we call prun inside this unittest but prun
> needs objects like xpdRE or so to run. In either beamline or local
> computer, we shouldn't need it because we will run icollection or ipyhton
> --profile=collection either way (and test on this behavior passed)
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/xpdAcq/xpdAcq/pull/138#issuecomment-197112897>
>



-- 
Prof. Simon Billinge
Applied Physics & Applied Mathematics
Columbia University
500 West 120th Street
Room 200 Mudd, MC 4701
New York, NY 10027
Tel: (212)-854-2918 (o) 851-7428 (lab)

Condensed Matter Physics and Materials Science Dept.
Brookhaven National Laboratory
P.O. Box 5000
Upton, NY 11973-5000
(631)-344-5661

email: sb2896 at columbia dot edu
home: http:// <http://nirt.pa.msu.edu/>bgsite.apam.columbia.edu/

chiahaoliu:
Yes, this way makes more sense than my previous approach. I will think along this direction.

chiahaoliu:
I tried to implement logic described above but the main challenges I had so far are:
  - real experiment objects like `pe1c` or `cs700` are created during startup process at XPD. So we need to set up attributes of `glbl` class, such as `glbl.area_det = pe1c` during this process, otherwise we can't have access to these real experiment objects unless we run the startup scripts directly, which is not preferable according to our discussion with Tom weeks ago. Therefore if we are about to obtain these objects during startup, we might inevitably need a logic inside `999-load.py` to decide whether we are running simulation or real experiment.

  - we also want to get rid of package dependencies of `bluesky` or `databroker`. The most obvious way I can think of is to import and set up those modules before going into `xpdAcq`, likely to be in `999-load.py`. Therefore, again, we need a logic to decide whether we are running simulation or real experiment so that we can provide imports accordingly

These two are the points have puzzled me a lot :(

chiahaoliu:
Maybe another possible approach is to have `998-init_simulation.py`  in ipython startup directory?
Contents of `998-init_simulation.py` could be:
```
if hostname != BEAMLINEHOSTNAME:
    from unittest.mock import MagicMock
    from xpdacq.mock_object import *
    pe1c = MagicMock()
    cs700 = MagicMock() 
    Count = MagicMock()
    ...
else:
     pass
```
In this sense, `999-load.py` can always grab appropriate objects depends on simulation or not and `glbl.py` will be much clean since those annoying conditional statements have been taken care during startup. This approach is slightly intrusive but it makes us easy to maintain our simulation environment in the future.

sbillinge:
I think we should just delete this pull-request now.  We have refactored globals and we can make a new branch and do the job of creating the new scan/xp object.  We should chat though because I am not 100% sure if we need both, or just an ``xp`` level object.

chiahaoliu:
I think we probably need either of them, as long as we create an object that is the `Union` of `sample.md` and `scan.md` before dumping into run engine. I will create a new PR starting from current `xpdAcq/master`

sbillinge:
I think the problem with this approach is that Union doesn't create a new
object, it just merges those two metadata dictionaries.

We have an ``Exposure()`` class. We could either move that out of the
``get_light_images()`` into ``prun``
or
Perhaps more appropriately we could change its name to ``Scan()`` class,
and move it into ``prun`` and change the ``xp_`` to ``sc_`` and maybe
change the current ``sc_`` md to ``sp_`` metadata.  This would make it all
more logical.

S

On Wed, Mar 16, 2016 at 11:39 PM, Timothy Liu <notifications@github.com>
wrote:

> I think we probably need either of them, as long as we create an object
> that is the Union of sample.md and scan.md before dumping into run
> engine. I will create a new PR starting from current xpdAcq/master
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/xpdAcq/xpdAcq/pull/138#issuecomment-197677922>
>



-- 
Prof. Simon Billinge
Applied Physics & Applied Mathematics
Columbia University
500 West 120th Street
Room 200 Mudd, MC 4701
New York, NY 10027
Tel: (212)-854-2918 (o) 851-7428 (lab)

Condensed Matter Physics and Materials Science Dept.
Brookhaven National Laboratory
P.O. Box 5000
Upton, NY 11973-5000
(631)-344-5661

email: sb2896 at columbia dot edu
home: http:// <http://nirt.pa.msu.edu/>bgsite.apam.columbia.edu/

PR #147, opened 2016-03-18
Add background scan

chiahaoliu:
Add `background()` for collecting background images.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -124,18 +124,18 @@ def _read_dark_yaml():
 def validate_dark(light_cnt_time, expire_time, dark_scan_list = None):
     ''' find the uid of appropriate dark inside dark_base
     
-        Parameters
-        ----------
-        light_cnt_time : float
-            exposure time of light image, expressed in seconds
-        expire_time : float
-            expire time of dark images, expressed in minute
-        dark_scan_list : list, optional
-            a list of dark dictionaries
-        Returns
-        -------
-        dark_field_uid : str
-            uid to qualified dark frame
+    Parameters
+    ----------
+    light_cnt_time : float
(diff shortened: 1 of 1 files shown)


xpdacq/xpdacq.py
```diff
@@ -244,6 +244,29 @@ def calibration(sample, scanplan, **kwargs):
     prun(sample, _scanplan)
     # this way is cleaner and dark is collected as well. but "no calibration file" warning might appear while people are doing calibration run.
 
+def background(sample, scanplan, **kwargs):
```
sbillinge: should this be ``_scanplan`` in the argument?

xpdacq/xpdacq.py
```diff
@@ -244,6 +244,29 @@ def calibration(sample, scanplan, **kwargs):
...
+            object carries scanplan metadata
+        **kwargs : dict
+             dictionary that will be passed through to the run-engine metadata
+        
+        Returns
+        -------
+        None
+    '''
+    scanplan = _scanplan
+    _scanplan.md.update({'sp_isbackground':True})
```
sbillinge: I think this should be ``'sc_isbackground'``, no? It is not an attribute of the scanplan.

sbillinge:
I can't pull any of these things in until we have tests for testing all the functionality in our xpdacq.py module.  I think it is important to work on that now that we have mocking working.

chiahaoliu:
Yes, I agree that is more important.

xpdacq/xpdacq.py
chiahaoliu: I create a copy of `scanplan` inside background function called `_scanplan` so that we can play with md without changing original `scanplan` object.

sbillinge: where do you do that?  And why do you do that?

PR #155, opened 2016-03-18
Fix scan class

chiahaoliu:
create a copy inside `Scan` by creating a new dict. Now confirm `ScanPlan.md` remains unchanged when we do changes on `Scan` class level.

### DIFF
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -365,17 +365,8 @@ class Scan(XPD):
         self.type = 'sc'
         self.sp = scanplan
         self.sa = sample
-        self.md = self.sp.md
+        self.md = dict(self.sp.md) # create a new dict copy.
         self.md.update(self.sa.md)
- #       self._yamify()    # no need to yamify this    
-
-class Xposure(XPD):
-    def __init__(self,scan):
-        self.type = 'xp'
-        self.sc = scan
-        self.md = self.sc.md
- #       self._yamify()    # no need to yamify this
-
 
 def export_data(root_dir=None, ar_format='gztar', end_beamtime=False):
     """Create a tarball of all of the data in the user folders.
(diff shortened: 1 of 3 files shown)


xpdacq/beamtime.py
```diff
@@ -365,17 +365,8 @@ def __init__(self,sample, scanplan):
         self.type = 'sc'
         self.sp = scanplan
         self.sa = sample
-        self.md = self.sp.md
+        self.md = dict(self.sp.md) # create a new dict copy.
```
sbillinge: still not sure about this.  Does it pass all the tests?

I still think it would be better to leave the ``sp`` and ``sa`` metadata in ``self.sp.md`` and so on, then just ``self.md = {}``

Then right before RE we can create a 
```
metadatadict = sc.md
metadatadict.update(sc.sp.md)
metadatadict.update(sc.sa.md)
xpdRE(..., metadatadict)
```
sthg like that.  We can do away with ``Union``, which I wrote but I don't think I like it now.

chiahaoliu: I got this idea from Tom's `collect_time_series` function [here](https://github.com/xpdAcq/xpdAcq/blob/master/xpdacq/xpdacq.py#L418-L419) and this logic pass unittest cases of 1) directly playing md in `Scan` class and 2) passing `Scan` to `prun` and `dark`.

I can change code to the way describe above so that it is more intuitive while reading.

sbillinge: I see.  Perhaps this is a better way.  I was afraid that the metadata was duplicated (items will be both in ``scan.md`` and in ``scan.sp.md`` for example, but I guess it is ok if we only pass ``scan.md`` to the run engine, which is I guess what happens, and then scan kind of dies and doesn't get yaml'd or anything.  Then the ``md`` is not duplicated in the exposure header.

PR #159, opened 2016-03-23
REF: refactor get_light_images as a plan

tacaswell:
This is going to break a bunch of things as I renamed a function.

This now lets you do things like

```python
RE(chain(get_light_images_plan(sc, **kwargs) for sc in list_of_scans))
```

or

```python
RE(chain(get_light_images(scan, exposure=j) for j in range(1, 15))
```

and all of the pause/resume will work as expected and suspenders will be more reliable.


chiahaoliu:
:+1:  the first UC 
``` 
RE(chain(get_light_images_plan(sc, **kwargs) for sc in list_of_scans)) 
```
might be really useful!

sbillinge:
Thanks Tom, that's really nice, but it doesn't quite get the use-case which is more like this: 

0) user opens (runscans) environment in a window

UC1
1) user types ``start()``
2) program loads default queue list from db
3) program pops the first plan in the list and runs it
4) scan finishes successfully.
5) program reads, possibly updated, queue list from db
6) program pops the first plan and runs it
7) repeat
8) program finds nothing more in the list and politely exits

UC2
as UC1.1 to UC1.3 then
4) user types CTL-C
5) program pauses scan and returns cmd line back to user
6) user fixes issue
7) user types ``resume()``
8) program resumes from last checkpoint, which may be a checkpoint within a scan, or a restart of the top scan in the (possibly, actually likely, updated) list
9) user types CTL-C again
10) user confused and needs to start over.  User types ``abort()``
11) program does something, but the most intuitive thing would be that it aborts both the current scan AND stops running the queue list.

For these UCs to work, either we need to (a) wrap ``RE()`` again (and reintroduce brittleness?), or (b) have RE take as an option an iterable object containing plans but with a fancy ``pop`` method.  Is it possible/desirable?

If you don't like (b) I would probably go with the brittleness (as we are now) until we get too many complaints from the users because it is too brittle.

tacaswell:
```python
queue_of_plans = deque(...)  # or a heap or a priority queue
def queue_plan():
    while len(queue_of_plans):
         plan = que_of_plans.popleft()
         yield from plan

RE(queue_plan)
```

sbillinge:
yes, that would do it, if popleft can read a potentially updated list (sorry, didn't encounter it before)

sbillinge:
Great.  Per conversation our conversation this lunchtime, we will go with our current (possibly brittle but seems to be working) solution for v0.3 release, but think about a solution like this for v0.4.

CJ-Wright:
Closed by #202

PR #164, opened 2016-04-13
Add background and dryrun function

chiahaoliu:
Add `background` and `dryrun` function and include `dryrun` option to `_unpack_and_run`. Code passed both simulation and `unittest`.

I still leave `get_light_image_dryrun` and other similar dryrun function in in `xpdaca.py`. If this refactoring is okay, I will clean them up in next commit.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -216,7 +216,7 @@ def _auto_load_calibration_file():
     config_md_dict = {'sc_calibration_parameters':config_dict, 'sc_calibration_file_name': os.path.basename(config_in_use), 'sc_calibration_file_timestamp':config_time}
     return config_md_dict
 
-def prun(sample, scanplan, auto_dark = glbl.auto_dark, **kwargs):
+def prun(sample, scanplan, auto_dark = None, **kwargs):
     ''' on this sample run this scanplan
 
     Parameters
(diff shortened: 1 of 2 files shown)


xpdacq/xpdacq.py
```diff
@@ -257,6 +257,48 @@ def calibration(sample, scanplan, auto_dark = glbl.auto_dark, **kwargs):
     _execute_scans(scan, auto_dark, auto_calibration = False, light_frame = True, dryrun = False)
     return
 
+def background(sample, scanplan, auto_dark = glbl.auto_dark, **kwargs):
```
sbillinge: Check if it is better to do it this way or to do ``auto_dark = None`` then inside the function
```
if auto_dark == None:
    auto_dark = glbl.auto_dark
```
I don't mind which way it is, but I thought that @tacaswell  liked it better this way....

xpdacq/xpdacq.py
```diff
@@ -287,54 +329,45 @@ def dark(sample, scanplan, **kwargs):
     _yamify_dark(dark_def)
     return dark_uid
 
-def dryrun(sample,scan,**kwargs):
-    '''same as run but scans are not executed.
+def background(sample, scanplan, **kwargs):
```
sbillinge: The function name and the docstring do not seem to agree with each other.  name suggests a background scan but docstring a dryrun

xpdacq/xpdacq.py
chiahaoliu: :+1:

xpdacq/xpdacq.py
chiahaoliu: This was corrected in a commit later than this one.

chiahaoliu:
`auto_dark` issue has been fixed in last commit.

PR #167, opened 2016-04-14
Refactor dk_window

chiahaoliu:
Make `dk_window` completely an attribute to `ScanPlan` object as discussed in issue 145. Also some cleaning on outdated function during refactoring.

Code now passes `unittest` and simulation mode.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -22,6 +22,7 @@ import numpy as np
 import copy
 import sys
 import uuid
+import warning
 from configparser import ConfigParser
 from xpdacq.utils import _graceful_exit, _RE_state_wrapper
 from xpdacq.glbl import glbl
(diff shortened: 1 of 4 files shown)


xpdacq/xpdacq.py
```diff
@@ -179,7 +179,7 @@ def _auto_dark_collection(scan):
     except KeyError:
         # protection, shouldn't happen
         warnings.warn("It seems your ScanPlan object wasn't instantiated properly."
-                        "This scan will keep going but please check your package after this scan")
+                        "This scan will keep going but please contact instrument scientist immediately")
```
sbillinge: This message is still not quite right.  Please make it this:

`` "It seems your ScanPlan object wasn't instantiated properly. This may indicate a problem with the current version of the code."
 -                        "This scan will keep going but please notify the instrument scientist who can post a bug report"

chiahaoliu: Thanks, this has been corrected in next commit

PR #174, opened 2016-04-21
working on documentation

chiahaoliu:
This is the first revision on documentation on `xpdacq.github.io`. Syntax is updated and align with current ``xpdAcq``. Next step is to finalize contents we want to release and ``autodoc`` to package.


chiahaoliu:
Replace with new PR. close

PR #178, opened 2016-04-25
Work documentation

chiahaoliu:
Finished another round of revision. This PR supposed to be in the same one as PR #174, but due to git confusion, I can't push to original branch. Only the last commit introduces new changes, compared to PR #174


chiahaoliu:
found it duplicate. Close.

PR #190, opened 2016-05-18
bluesky pass-through v1

chiahaoliu:
This PR is to show one approach of passing arbitrary bluesky plan to ``prun``
Current API is like:
````
from bluesky.plans import Count
count  =  Count([pe1c, em, tth_cal], 5) # an arbitrary scan

prun(2, count, bsky_plan = True) # force user to be clear

# if user forgot to specify:
prun(2, count)
TypeError: Incorrect object assignment on <class 'xpdacq.beamtime.ScanPlan'>.
Remember xpdAcq like to think "run this Sample(sa) with this ScanPlan(sp)"
Please do bt.list() to make sure you are handing correct object type
````
code passes ``unittest`` and simulation mode. To run unittest, real bluesky package is needed.


chiahaoliu:
Decide to take another approach at PR #191, closed.

PR #191, opened 2016-05-19
bluesky pass-through v2

chiahaoliu:
This PR shows another approach of bluesky pass-through. Assignment happens at ``ScanPlan`` level. Current API is like:
````
from bluesky.plans import Count
c = Count([pe1c], 2)

ScanPlan('my_bluesky', 'bluesky', bsky_plan = c) 
# bsky_plan is necessary as we have many other positional arguments 

prun(2, 11) # index 2 = 'l-user' sample, index 11 = 'my_bluesky' ScanPlan 
````

This branch passed simulation but ``unittest`` hasn't been written. We can compare this approach with another version and choose which one to finalize.


sbillinge:
@chiahaoliu this is better but it doesn't make any sense to me why you wouldn't just pass the bluesky plan in the dictionary with the scanplan parameters.  This way you don't change the interface.

Right now, it looks like:
``ScanPlan('name','type',{scanplan-parameter-dict})``
plus some optional args.  So it seems straightforward that when `type` is `bluesky` then the scan-plan-parameter dict contains the bluesky Plan object and unpack_and_run just passes that object through somehow?  

What is the need for the ``bsky_plan``?

chiahaoliu:
Yes, I think this approach is more reasonable than assign ``bsky_plan``. I came out ``bsky_plan`` argument as I wanted to independently assign an object to ``ScanPlan`` but it turns out your approach is more consistent with xpdAcq API. I will finish this PR with refactored API.

Current API is like:
````
In [1]: from bluesky.plans import Count

In [2]: from bluesky.examples import det, motor

In [3]: cc = Count([det],2)

In [4]: ScanPlan('bs','bluesky',{'bluesky_plan':cc})
INFO: You are handing a "bluesky" type scan.
Please go to https://nsls-ii.github.io/bluesky/plans.html
for complete guide on on how to define a plan.
INFO: This ScanPlan does not support auto-dark subtraction
scanplan_params = {'bluesky_plan': <bluesky.plans.Count object at 0x10c91d630>}
Out[4]: <xpdacq.beamtime.ScanPlan at 0x103afd400>

In [5]: prun(2,9)
INFO: No calibration file found in config_base. Scan will still keep going on
WARNING: There is no wavelength information in your sample acquire object
```` 

code passed simulation and ``unittest`` will be finished pretty soon.

chiahaoliu:
Code passed simulation and ``unittest`` at this point

sbillinge:
OK, merging in.  There is a small type in line 412 of beamtime.py error message (two on's and a missing space), but I think everything looks good so I am pulling this in.  Please fix the typo in a future pull request.

PR #198, opened 2016-06-17
Update setup.py

chiahaoliu:
This PR is to update information used in setup.py. It updates url pointing to xpdAcq and pack ``bluesky`` with it. Now user can do ``python setup.py develop`` after clone ``xpdAcq``. ``bluesky`` and its dependencies will be installed as well.


CJ-Wright:
Do we want to use [versioneer](https://github.com/warner/python-versioneer) to handle the version numbers? I believe this is what the NSLS-II stack does, right @danielballan?

CJ-Wright:
Superseded by lots of other PRs

PR #209, opened 2016-09-18
ENH: import sample

chiahaoliu:
Enhance ``import_sample`` functionality. 

This PR mainly focus on increasing speed, now it takes ~700ms for a list of 80 Sample objects.

Current logic of ``bt.list`` is:
1. when calling ``import_sample`` function, ``Sample`` object list is entirely flushed and updated based on current spreadsheet contents. ``ScanPlan`` object list stays the same.

2. when creating objects in command line, like ``Sample(.....)``, objects in ``bt.list`` with the same sample name will be updated(overwritten), while keep index unchanged.

### DIFF
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -321,8 +321,9 @@ class Beamtime(ValidatedDictLike, YamlDict):
             self.scanplans.remove(old_obj)
             self.scanplans.insert(old_obj_ind, scanplan)
         # yaml sync list
-        self._referenced_by.extend([el for el in self.scanplans if el
-                                    not in self._referenced_by])
+        #self._referenced_by.extend([el for el in self.scanplans if el
+        #                            not in self._referenced_by])
+        self._referenced_by.append(scanplan)
 
     @property
     def md(self):
@@ -355,8 +356,9 @@ class Beamtime(ValidatedDictLike, YamlDict):
             self.samples.remove(old_obj)
             self.samples.insert(old_obj_ind, sample)
         # yaml sync list
-        self._referenced_by.extend([el for el in self.samples if el
-                                    not in self._referenced_by])
+        #self._referenced_by.extend([el for el in self.samples if el
+        #                            not in self._referenced_by])
+        self._referenced_by.append(sample)
 
     @classmethod
     def from_yaml(cls, f):
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -361,6 +361,8 @@ class Beamtime(ValidatedDictLike, YamlDict):
         # filtering logic is handle when importing sample
         #self._referenced_by.extend([el for el in self.samples if el
         #                            not in self._referenced_by])
+        # simply append object to list to increase speed
+        # filtering logic is handle when importing sample
         self._referenced_by.append(sample)
 
     @classmethod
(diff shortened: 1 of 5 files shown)


CJ-Wright:
This looks ok to me, although I am not so certain about what this is doing exactly.
You might look at the coverage to see what lines were not hit by the tests

xpdacq/beamtime.py
```diff
@@ -243,18 +245,17 @@ def register_sample(self, sample):
...
                        sample.get('sample_name')].pop()
             old_obj_ind = self.samples.index(old_obj)
             self.samples.remove(old_obj)
             self.samples.insert(old_obj_ind, sample)
         # yaml sync list
-        self._referenced_by.extend([el for el in self.samples if el
-                                    not in self._referenced_by])
+        # simply append object to list to increase speed
+        # filtering logic is handle when importing sample
+        self._referenced_by.append(sample)
```
tacaswell: should `_referenced_by` be a `set`?

tacaswell:
This looks like you really want an ordered dictionary as your data structure?

xpdacq/beamtime.py
chiahaoliu: :+1:  I would need to think about implementation details

chiahaoliu:
@tacaswell can you elaborate more? Thanks!

Basically we wish to have a internal list (``_referenced_by`` in this PR)  that belongs to ``bt`` object. The internal list servers a reference to update objects and its corresponding yaml files when information in ``bt`` is updated, 

In addition to aforementioned property, we also want this list to respect chronological logic, object created in later time should appear in the bottom when we print this list (therefore, ``set`` might not be appropriate)

sbillinge:
I see, we are revisiting the serialization issue. 

Is this just for `sample` attributes, coming from the csv?  In that case, I thought we had a more simple solution  of just overwriting all yamls when a csv is uploaded?  this doesn't need any an list.

@chiahaoliu please could you give one or two examples of how your proposed structure will be used?

chiahaoliu:
Current structure already allows us to have exactly the work-flow described above- resync every time ``bt`` info is updated and respect chronological order- with an acceptable performance (700 ms for 80 samples).

I was wondering if there is  a better way of implementing whole structure, even though current structure seems to be okay (thanks @danielballan for customized ``yamldict`` class) 

Here is a demo of current structure:

````
In [4]: bt.samples[0].md
Out[4]:
{'bt_experimenters': ['Tim', 'Liu'],
 'bt_piLast': 'Alva',
 'bt_safN': '300056',
 'bt_uid': '93b1c484',
 'bt_wavelength': 0.184649,
(....too long to display)

In [5]: bt.md
Out[5]:
{'bt_experimenters': ['Tim', 'Liu'],
 'bt_piLast': 'Alva',
 'bt_safN': '300056',
 'bt_uid': '93b1c484',
 'bt_wavelength': 0.184649}

In [6]: bt.update({'bt_piLast':'new_field'})

# propagate down to all samples
In [7]: bt.samples[0].md
Out[7]:
{'bt_experimenters': ['Tim', 'Liu'],
 'bt_piLast': 'new_field',
 'bt_safN': '300056',
 'bt_uid': '93b1c484',
 'bt_wavelength': 0.184649
(....too long to display)

In [8]: bt.samples[1].md
Out[8]:
{'bt_experimenters': ['Tim', 'Liu'],
 'bt_piLast': 'new_field',
 'bt_safN': '300056',
 'bt_uid': '93b1c484',
 'bt_wavelength': 0.184649,
(....too long to display)
````

danielballan:
IIRC you don't even need to type `.md`. It's there for back compat, but `bt.samples[1]` *itself* is dict-like.

PR #224, opened 2016-09-30
V0.5.1 patch

chiahaoliu:
This PR include two parts:

1. clean up old and messy simulation logic

  as per @tacaswell, software backend changes at beamline, it's better to take whatever is configured during start profile as it will be most updated one. Now local test/simulation environment is completely configured by ``start_simulation`` function

2.  correct workflow of ``xpdAcq``

  per @sbillinge, correct currently wrong workflow. Now mask is created during ``run_calibration`` and it persists as part of metadata until ``run_calibration``  or ``get_mask`` function is run again.

mask logic inside integration will be fixed separately at ``xpdAn`` side.


xpdacq/999-load.py
```diff
@@ -5,11 +5,29 @@
...
 
-if not glbl._is_simulation:
-    glbl.area_det = pe1c
-    glbl.shutter = shctl1
-    glbl.temp_controller = cs700
-    # let NameError handle missing object
+# experiment device being used in current plan
+try:
+    device_list = [pe1c, shctl1, cs700]
+    attribute_name = ['area_det', 'shutter', 'temp_controler']
```
sbillinge: typo, please clean

xpdacq/calib.py
```diff
@@ -78,6 +79,117 @@ def run_calibration(exposure=60, calibrant_file=None, wavelength=None,
...
+          .format("http://xpdacq.github.io/usb_Running.html#calib-manual"))
+    print()
+    # calibration, return a azimuthal integrator
+    ai = calibration(img, calibrant_file=calibrant_file,
+                     calib_collection_uid=calib_collection_uid,
+                     wavelength=wavelength, detector=detector,
+                     gaussian=gaussian)
+
+    # masking
+    print('INFO: create mask')
```
sbillinge: We should be careful here.  The correct workflow is that the mask is created "during the calibration workflow".  There are two possibilities:
1) The automask from nickel (or whatever calibrant), plus the edge mask, plus (maybe) the polygon mask for the beamstop mount,  is adequate for all samples.  
2) We need to collect data from a weakly scattering sample such as an empty kapton tube, run automask on that to get the auto-mask part of the mask, then create the mask as automask-mask+edge-mask(+beamstop mask).

If (1) is correct then the workflow coded here is acceptable.  If we need to use (2) then this is not acceptable.

Either way, it may be better to have a separate mask-creation logic that is run outside of `run_calibration`.  It is generally better that functions just do what they appear to do.  So `run_calibration()` does just that...it runs the get-calibration logic.  Then we have a `run_mask_builder()` function that does that....it runs the automask and builds the mask. 

In this way, we can leave it to the user what sample they want to mount to get the automask they want.   The documentation will then have to reflect this.  The sequence could be:
1) load Ni
2) `run_calibration()`
3) load empty kapton
4) `run_mask_builder()`
5) load sample
6) `prun(sampleObject,scanPlanObject)`
7)  `view_last_image()`
8)  toggle mask on and off in `XPDsuite` to make sure that the auto-built mask is acceptable
9)  all good?  start experiment.....

xpdacq/calib.py
```diff
@@ -78,6 +79,117 @@ def run_calibration(exposure=60, calibrant_file=None, wavelength=None,
...
+
+    See also
+    --------
+    xpdan.tools.mask_img
+    """
+    mask = mask_img(img, geometry_object, **mask_dict)
+    print("INFO: add mask to global state")
+    glbl.mask = mask
+
+    if save_name is not None:
```
sbillinge: if `save_name` is None, what happens?  No mask saved?

xpdacq/calib.py
```diff
@@ -122,35 +122,103 @@ def run_calibration(exposure=60, dark_sub=True, calibrant_file=None,
...
+    # save yaml dict used for xpdAcq
+    yaml_name = glbl.calib_config_name
+    with open(os.path.join(glbl.config_base, yaml_name), 'w') as f:
+        yaml.dump(glbl.calib_config_dict, f)
 
     return ai
 
 
-def get_mask(img, geometry_object, mask_dict, save_name=None):
+def run_mask_builder(exposure=60, dark_sub=True,
```
sbillinge: check with max what is an appropriate default exposure for empty kapton.  it might be much longer than this.

xpdacq/calib.py
chiahaoliu: yes, no file will be saved, only glbl attribute is updated. I put this behavior as we are going to save the mask when we run ``integrate_and_save`` on ``xpdAn`` side.

CJ-Wright:
@chiahaoliu I merged the compression stuff into the An side so the tests are passing.

xpdacq/calib.py
```diff
@@ -122,35 +122,103 @@ def run_calibration(exposure=60, dark_sub=True, calibrant_file=None,
...
+        dark_uid = light_header.start['sc_dk_field_uid']
+        dark_header = glbl.db[dark_uid]
+        dark_img = np.asarray(glbl.db.get_images(dark_header,
+                                glbl.det_image_field)).squeeze()
+    for ev in glbl.db.get_events(light_header, fill=True):
+        img = ev['data'][glbl.det_image_field]
+        if dark_sub:
+            img -= dark_img
+
+    print("INFO: masking you image ....")
```
sbillinge: Please delete this message.  It is kind of obvious that a mask is being created when you run ``mask_builder``

PR #227, opened 2016-10-03
Doc web

chiahaoliu:
PR to share documentation edits


CJ-Wright:
Closes xpdAcq/xpdAn#22

sbillinge:
might want to reopen this issue, or a new one...I don't think there is user documentation for the auto-mask capability, though its use is there in the workflow.

This should definitely be in the An side of things though, so if we create a separate xpdAn user-doc (which I think we need) it should go in there, and attached to v0.6 release.

PR #246, opened 2016-10-06
V0.5.0 test

chiahaoliu:
modified code after testing at XPD


xpdacq/glbl.py
```diff
@@ -30,7 +30,7 @@
     BASE_DIR = os.getcwd()
 else:
     #BASE_DIR = os.path.expanduser('~/')
-    BASE_DIR = os.path.abspath('/direct/XF28ID1/pe1_data')
+    BASE_DIR = os.path.abspath('/direct/XF28ID1/pe1_data/UserArea/XPDhome/xpdUser')
```
sbillinge: this should not have xpdUser on it

xpdacq/glbl.py
```diff
@@ -29,7 +29,8 @@
 if simulation:
     BASE_DIR = os.getcwd()
 else:
-    BASE_DIR = os.path.expanduser('~/')
+    #BASE_DIR = os.path.expanduser('~/')
+    BASE_DIR = os.path.abspath('/direct/XF28ID1/pe1_data/UserArea/XPDhome/xpdUser')
```
sbillinge: no xpdUser here

xpdacq/utils.py
```diff
@@ -277,7 +277,7 @@ def load(self, saf_num):
 
         self.pd_dict = pd.read_excel(os.path.join(self.src_dir,
                                                   xl_f.pop()),
-                                     skiprows=0)
+                                     skiprows=[1])
```
sbillinge: I will assume it is correct, but please check.  Does this skip row 1 or all rows up to 1?  I would rather expect to see `skiprows=[:1]`

xpdacq/glbl.py
chiahaoliu: updated with commit just pushed

xpdacq/utils.py
chiahaoliu: we need first row to be a column name so I only skip second row

CJ-Wright:
You need to propagate the refactor of `import_sample` -> `import_sample_info` into the tests.

chiahaoliu:
will separate in small PR

CJ-Wright:
Awesome, thanks

CJ-Wright:
@chiahaoliu Do you mind rebasing onto master if this PR is still going to be used.

chiahaoliu:
@CJ-Wright this PR will not be used, instead, it is served as bookkeeping purpose. As described above, I will break it into several small ones. Now I am 2 small PR away from completing this PR, if you don't mind, can I leave it here? Of course, I am happy to rebase as well.

CJ-Wright:
Can we close this then?

chiahaoliu:
close as it will be separated into small PRs

PR #276, opened 2016-10-13
make wavelength an optional argument in start_beamtime

chiahaoliu:


xpdacq/calib.py
```diff
@@ -143,7 +143,7 @@ def run_calibration(exposure=60, dark_sub_bool=True, calibrant_file=None,
 
 
 def run_mask_builder(exposure=300, dark_sub_bool=True,
-                     polarization_factor=0.99,
+                     poloarization_factor=0.99,
```
sbillinge: This looks like a typo....why are you doing this change?

chiahaoliu:
reissue with new PR

PR #277, opened 2016-10-13
TST/CI: add conftest to provide a databroker and move coverage to onl…

CJ-Wright:
…y if pass


CJ-Wright:
I think that this PR needs to be re-thought, especially as this overlaps with some xpdSim functionality.

PR #281, opened 2016-10-14
Pre populate scanplan during start_beamtime

### LINKED ISSUE #280: default setup ScanPlan ([0]) should be 5s
break issues from #247 into small parts

chiahaoliu:
fix #280

### DIFF
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -38,9 +38,12 @@ def _start_beamtime(PI_last, saf_num, experimenters=[], *,
         src = os.path.join(os.path.dirname(__file__), 'Ni24.D')
         dst = os.path.join(glbl.usrAnalysis_dir, 'Ni.D')
         shutil.copy(src, dst)
-        # import sample
-        # FIXME: leave for user?
-        # import_sample(bt, saf_num)
+        
+        # pre-populated scan plan
+        ScanPlan(bt, ct, 60)
+        ScanPlan(bt, ct, 30)
+        ScanPlan(bt, ct, 10)
+
         return bt
 
 
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -41,6 +41,8 @@ def _start_beamtime(PI_last, saf_num, experimenters=[], *,
 
         # pre-populated scan plan
         ScanPlan(bt, ct, 5)
+        ScanPlan(bt, ct, 1)
+        ScanPlan(bt, ct, 10)
         ScanPlan(bt, ct, 30)
         ScanPlan(bt, ct, 60)
 
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -38,11 +38,11 @@ def _start_beamtime(PI_last, saf_num, experimenters=[], *,
         src = os.path.join(os.path.dirname(__file__), 'Ni24.D')
         dst = os.path.join(glbl.usrAnalysis_dir, 'Ni.D')
         shutil.copy(src, dst)
-        
+
         # pre-populated scan plan
-        ScanPlan(bt, ct, 60)
+        ScanPlan(bt, ct, 5)
         ScanPlan(bt, ct, 30)
-        ScanPlan(bt, ct, 10)
+        ScanPlan(bt, ct, 60)
 
         return bt
 
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -41,6 +41,7 @@ def _start_beamtime(PI_last, saf_num, experimenters=[], *,
 
         # pre-populated scan plan
         ScanPlan(bt, ct, 5)
+        ScanPlan(bt, ct, 0.1)
         ScanPlan(bt, ct, 1)
         ScanPlan(bt, ct, 10)
         ScanPlan(bt, ct, 30)
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -10,6 +10,9 @@ from .glbl import glbl
 from .beamtime import *
 from .utils import _graceful_exit
 
+# list of exposure times for pre-poluated ScanPlan inside
+# _start_beamtime
+EXPO_LIST = [5, 0.1, 1, 10, 30, 60]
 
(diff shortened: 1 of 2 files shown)


xpdacq/beamtimeSetup.py
```diff
@@ -38,9 +38,12 @@ def _start_beamtime(PI_last, saf_num, experimenters=[], *,
         src = os.path.join(os.path.dirname(__file__), 'Ni24.D')
         dst = os.path.join(glbl.usrAnalysis_dir, 'Ni.D')
         shutil.copy(src, dst)
-        # import sample
-        # FIXME: leave for user?
-        # import_sample(bt, saf_num)
+        
+        # pre-populated scan plan
+        ScanPlan(bt, ct, 60)
```
sbillinge: do we have a ct 5?  I think we use that in the documentation.  I would advocate for a ct 1 also, and possibly a ct "min".  Why?  ct "min" gives people the shortest time that is available.  In this case "min" is 1/frame-rate, i.e., acquisition time.

btw, what will happen if an exposure is requested that is shorter than the 1/frame-rate ?  Do we have exception handling for that case?  The right behavior in general should be to not create it and to throw a warning.  However, the right behavior in the startup sequence should probably be the same, but maybe to pause with a question to the user to make sure that the frame-rate is set correctly.

xpdacq/beamtimeSetup.py
```diff
@@ -38,11 +38,11 @@ def _start_beamtime(PI_last, saf_num, experimenters=[], *,
         src = os.path.join(os.path.dirname(__file__), 'Ni24.D')
         dst = os.path.join(glbl.usrAnalysis_dir, 'Ni.D')
         shutil.copy(src, dst)
-        
+
         # pre-populated scan plan
-        ScanPlan(bt, ct, 60)
+        ScanPlan(bt, ct, 5)
```
sbillinge: ah, this may have just answered my first question....

chiahaoliu:
@sbillinge can you have a look if this PR still needs edits? thanks

xpdacq/tests/test_beamtimeSetup.py
```diff
@@ -94,9 +94,9 @@ def test_start_beamtime(self):
         self.assertEqual(self.experimenters, self.bt.get('bt_experimenters'))
         self.assertEqual(self.wavelength, self.bt.get('bt_wavelength'))
         self.assertEqual(os.getcwd(), self.home_dir)
-        # FIXME
-        # strtScnLst = ['bt_bt.yml','ex_l-user.yml','sa_l-user.yml','sp_ct_0.1.yml','sp_ct_0.5.yml','sp_ct_1.yml','sp_ct_5.yml','sp_ct_10.yml','sp_ct_30.yml']
-        # self.assertEqual(newobjlist,strtScnLst)
+        # test prepoluate ScanPlan
+        self.assertEqual(len(self.bt.scanplans), 3)
+        self.assertEqual(self.bt.scanplans[0]['sp_args'], (5,))
```
sbillinge: this test is ok, but it is not testing much of the functionality is it?

Also, I think we wanted  5s, "min", 1s,  10s, 30s, 60s in the prepopulated list, something like that (with 5s in the first place)?

CJ-Wright: Would it be more functional to do this assert in a for loop?
```python
expected_sp_args = [5, 60, 1, 10, 30, 60]
for plan, expected in zip(self.bt.scanplans, expected_sp_args):
    self.assertEqual(plan['sp_args'], expected)
    # Maybe to a plumbing test running a prun of the scan plan?
```
This way if we add a new scan plan we just need to change the `sp_args` to include the updated scan plan arguments.

chiahaoliu: @sbillinge functionality of ``ScanPlan`` objects are tested in ``test_beamtime.py`` (it is generated by exactly the same syntax and I test whether it can be created, updated when ``bt`` is changed, reloaded and run by RunEngine). Is there any suggested aspect I should cover?

for ``min`` argument, I need to implement it carefully as this might relate to the API of ``ScanPlan``.  Currently ``ct_0.1`` and ``ct_10``  in ``bt.list()`` is generated by inspecting function argument which is a float. I would prefer to separate it as another PR to ScanPlan as well.

chiahaoliu: @CJ-Wright thanks for the magic, I will include it into test as well

sbillinge: I think the what we want to test here, if possible, is that the right scanplans were pre-created successfully, not that they work successfully per se.

Based on cj's `expected_sp_args` list it looks as if `ct_60` is created twice, so that test should fail....

If "min" behavior is tricky, let's leave it at `0.1` instead of `min`, but put the `min` behavior in v0.6 release as an issue.  Here we are fixing something that is not broken.....though it will be easier to fix it soonish before it is broken.

Summary:
1) create issue on v0.6 for `min` behavior
2) finalize the list of pre-created ct's
3) code them (basically done)
4) code the tests to check they were successfully created (maybe read the yml's to make sure they were written or something?)
5) pull in and close the issue.

chiahaoliu:
okay, issue for ``min`` has been created and more tests about ``ScanPlan`` objects created during ``start_beamtime`` are added. Ready to review.

xpdacq/tests/test_beamtimeSetup.py
```diff
@@ -94,9 +94,19 @@ def test_start_beamtime(self):
...
+            self.assertEqual(sp['sp_args'], (expect_arg,))
+        # test if yml files are saved properly
+        for expo in EXPO_LIST:
+            f_path = os.path.join(glbl.scanplan_dir,
+                                  'ct_{}.yml'.format(expo))
+            self.assertTrue(os.path.isfile(f_path))
+        # test if it can be reloaded
+        for current_sp in self.bt.scanplans:
+            reload_sp = ScanPlan.from_yaml(current_sp.to_yaml())
+            self.assertEqual(reload_sp, current_sp)
```
CJ-Wright: Is there a way to check that `reload_sp` is not the exact same object as `current_sp`? This way we can make certain we loaded it up from the file and didn't short circuit anywhere. (I think hash magic might sort this out, maybe)

chiahaoliu: do with ``id`` to differentiate memory address

PR #289, opened 2016-10-18
REF: change prun over to xrun

### LINKED ISSUE #255: Yikes, do we want to change prun to xrun!
It is a pain that we have inadvertently used a python built-in function for our main scan-engine.  Also, it is called `prun` to mean `production run` but the logic has now changed so that we can run setup scans and background scans and whatever using this same run-engine, so it doesn't make any sense to call it `prun`.  If we are going to change it, now is the time!

CJ-Wright:
This is a search and destroy replacement of all prun instances to xrun.
Closes #255

### DIFF
--- a/doc/babble.rst
+++ b/doc/babble.rst
@@ -25,16 +25,16 @@ Start a beamtime
   * ``experimenters`` field is expected to be [‘first_name’, ‘last_name’, ‘fist_name’, ‘last_name’, ….]
 
 
-link bt to prun
+link bt to xrun
 """""""""""""""
 
 .. code-block:: python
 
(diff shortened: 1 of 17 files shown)


chiahaoliu:
@CJ-Wright awesome, I am doing the same PR, except for I include other small fix. Would you mind me rebasing your branch?

CJ-Wright:
@chiahaoliu Feel free to do whatever you want! just let me know if you want this pulled or are going to supersede it.

chiahaoliu:
looks good to me and code passes tests (plus it updates the documentation 👍)

PR #297, opened 2016-10-18
clean old documentation

chiahaoliu:
@sbillinge  @CJ-Wright 

This branch contains documentation with up-to-date code output. If the main body looks fine, I would like it to be reviewed and pull in than I could assign small parts of documentation to our group memebers for final review, thanks.


sbillinge:
Didn't get all through, but language will be tweaked so we can fix small issues then, so I am pulling in.

PR #307, opened 2016-10-20
DOC: add user scripts in quick_start

### LINKED ISSUE #229: finish user scripts in quick-start documentation

chiahaoliu:
fix #229

### DIFF
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -278,11 +278,11 @@ and assign it to the object we have called ``mybsplan`` in this example:
   from bluesky.plans import list_scan
 
   # it is entirely optional to add metadata to the scan, but here is what you would do:
-  mymd = {'memoy_aid': 'This metadata should be about the scan, not the sample which would be added when the scanplan is run',
+  mymd = {'memory_aid': 'This metadata should be about the scan, not the sample which would be added when the scanplan is run',
           'author': 'Simon',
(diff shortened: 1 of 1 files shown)


doc/quickstart.rst
```diff
@@ -526,20 +526,76 @@ User scripts
 Your experiment commands can be sequenced into scripts,
 to be executed one after the other as you desire.  To set this up, write a sequence of commands into a text file,
 save it with the extension ``.py`` in the ``userScripts`` directory with a memorable name, like ``myNightShiftScript.py``.
-Double and triple check your script, then when you are ready to execute it, in ``ipython`` session type:
 
+Here is an example for a script that runs two grid scans with different mesh sizes.
```
sbillinge: this should not be in quickstart, it is way too complicated, and using bluesky built-in plans.  Let's leave the quickstart as it was and we should discuss what we want to do in the way of scripts that users can use.

We can move this level of scripting to v0.6, and maybe old-style scripting that works but is not so robust, but easier to understand,  to v0.5patch.

Of course users can write scripts like this if they read bluesky documentation, but I don't want them seeing this in our xpdAcq documentation.

chiahaoliu: ok, sounds good

chiahaoliu:
close per discussion

PR #313, opened 2016-10-26
Versioneer

CJ-Wright:


chiahaoliu:
Looks good to me. @CJ-Wright  is there any other know-how we should know with versioneer?

CJ-Wright:
Going to use rever

Issue #292, opened 2016-10-18
reload glbl options after exiting out ipython


chiahaoliu:
as from #145
we would like to have following behavior:
1. reload `glbl` option everytime user exits and comes back to ipython session
2. `glbl` option should not be persistent over beamtimes

CJ-Wright:
We might also want to have a default global that the user can say "I no longer like all my changes to the global please just give me the defaults back".

PR #317, opened 2016-11-01
refactor glbl class

chiahaoliu:
This is a PR to share idea, not aiming for pulling at this stage (syntax updates are needed if we are heading this direction).

This refactoring would introduce following features

1. ``glbl`` now behaves like ``Beamtime`` object. Every time allowed key is updated, it automatically serializes the change so that makes reload of ``dark_window`` and other configurations possible. (address #292)

2.  As ``glbl`` now behaves like ``Beamtime``, it is also possible to ``an_glbl`` behaves like ``Sample`` objects (i.e. make them as a ``ChainMap``). In this way, we can reproduce the functionality of propagating updates to both objects once content from either of them is updated. 
[xpdAn issue 30](https://github.com/xpdAcq/xpdAn/issues/30) 

3. prevent user from setting arbitrary attributes, like ``glbl.apple = orange``(which is currently possible)

4. We can also make ``glbl`` as a ``ChaiMap`` with another ``dict`` contains required device objects. So we can potentially adopt the suggestion mentioned in 
[xpdAn issue 29](https://github.com/xpdAcq/xpdAn/issues/29), then we reduce the possibility of maintaining contents inside startup directory.

Any thught or suggestions?  @tacaswell @danielballan @sbillinge @CJ-Wright 

Thanks

### DIFF
--- a/xpdacq/glbl.py
+++ b/xpdacq/glbl.py
@@ -114,18 +114,17 @@ glbl_dict = {
              }
 
 
-# dict to store all necessary dict
+# dict to store all necessary objects
 xpd_device = {}
 
-def setup_xpdAcq(device_list, *, area_det, shutter, temp_controller, db,
-                 **kwargs):
+def setup_xpdAcq(*, area_det, shutter, temp_controller, db, **kwargs):
     """ function to set up required device/objects for xpdacq """
-    REQUIRED_KEYS = ['area_det', 'shutter', 'temp_controller', 'db']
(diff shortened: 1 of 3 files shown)


chiahaoliu:
Btw, with changes share in this PR, ``glbl`` is able to automatically serialize change and be reloaded from local ``yaml`` file.

chiahaoliu:
close based discussion. Revisit it later.

PR #320, opened 2016-11-02
Fix bt.list

### LINKED ISSUE #233: change behavior of bt.list()
As suggest of @tacaswell, better to change it to `OrderDict` but not a list

also have to take care of speed

Don't destroy existing `Sample` objects, we want to update but not destroy all and recreate.

persistent `ssetup` sample object at posiion 0

chiahaoliu:
closes #233 

Turn ``bt.samples`` and ``bt.scanplans`` into ``OrderedDict`` so that metadata will only be updated, not refreshed when redo ``import_sample_info``.

Also added sorting logic to reload sequence. In this way, we can make sure indices in ``bt.list`` remain unchanged after exit and come back to ipython

### DIFF
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -338,12 +338,12 @@ def _clean_info(obj):
     """ stringtify and replace space"""
     return str(obj).strip().replace(' ', '_')
 
-class MDOrderedDict(OrderedDict):
-    def get_md(self, ind):
-        """ special method to get metadata of a objects inside a dict """
-        obj_list = list(self.values())
-        md_dict = dict(obj_list[ind])
-        return md_dict
+
+class SliceableOrderedDict(OrderedDict):
+    def get_slice(self, ind):
+        """method to return value of OrderedDict based on index"""
+        val_list = list(self.values())
+        return dict(val_list[ind])
 
 
 class Beamtime(ValidatedDictLike, YamlDict):
@@ -392,8 +392,8 @@ class Beamtime(ValidatedDictLike, YamlDict):
                          bt_experimenters=experimenters,
                          bt_wavelength=wavelength, **kwargs)
         self._wavelength = wavelength
-        self.scanplans = MDOrderedDict()
-        self.samples = MDOrderedDict()
+        self.scanplans = SliceableOrderedDict()
+        self.samples = SliceableOrderedDict()
         self._referenced_by = []
         # used by YamlDict when reload
         self.setdefault('bt_uid', new_short_uid())
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -431,7 +431,7 @@ class Beamtime(ValidatedDictLike, YamlDict):
         # yaml sync list
         self._referenced_by.append(scanplan)
         # save order
-        with open(os.path.join(glbl.config_base,
+        with open(os.path.join(glbl['config_base'],
                                '.scanplan_order.yml'),'w+') as f:
             scanplan_order = {}
             for i, name in enumerate(self.scanplans.keys()):
@@ -448,7 +448,7 @@ class Beamtime(ValidatedDictLike, YamlDict):
         # yaml sync list
         self._referenced_by.append(sample)
         # save order
-        with open(os.path.join(glbl.config_base,
+        with open(os.path.join(glbl['config_base'],
                                '.sample_order.yml'),'w+') as f:
             sample_order = {}
             for i, name in enumerate(self.samples.keys()):
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -430,6 +430,15 @@ class Beamtime(ValidatedDictLike, YamlDict):
         self.scanplans.update({scanplan_name: scanplan})
         # yaml sync list
         self._referenced_by.append(scanplan)
+        # save order
+        with open(os.path.join(glbl.config_base,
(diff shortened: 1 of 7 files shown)


chiahaoliu:
@sbillinge @CJ-Wright this PR is ready for review

chiahaoliu:
No, this doesn't change the interface in terms of creation of beamtime objects. 

The only change is user used to to do `` bt.samples[1].md`` to interrogate metadata, now they would need to do ``bt.samples.get_md(1)`` instead.

If this PR/interface change is passed, I will create another issue of documentation to capture this change.

danielballan:
Fly-by comment as I quickly catch up on things here: I think `bt.sample[1]` is sufficient. We made `bt.samples[1].md` a synonym just for back-compatibility, if I recall correctly. I would say `[1]` is better than `.get_md(1)` because it is briefer and does not require knowing xpdacq-specific interface -- just the standard Python slicing syntax.

chiahaoliu:
Yes, I agree ``get_md`` is a bit too specific but it is a consequence of solving a problem we are facing. Let me explain the whole problem, maybe you can shed some light :)

The problem we are trying to solve here is that we want a logic as :
"When creating ``Sample`` object, If there is a ``Sample`` object has the same ``sample_name``, then we will update its content, otherwise we will create and register a new one. "

Before this PR, ``bt.samples`` is a list, so we could do standard slicing like ``bt.sample[1]``, however, when I tried to use list comprehension and ``index`` method to implement the logic above, I encountered a speed issue. When importing more than 100 samples in one time, it took more than 5mins to finish.

Based on this fact, I make ``bt.samples`` an ``OrderedDict`` with ``sample_name`` as the key and ``Sample`` object as the value in this PR so that we can use ``update`` method to achieve the logic. However, this implementation prohibits index-based slicing. That's why I make a special method of ``get_md(ind)`` to have similar interface as before. (or course, they can still do ``bt.samples['mySampleName']``, but I guess that will cause complains :p)

sbillinge:
Yes, I have been talking to Tim about this issue and we really need that we can update-and-extend the sample list rather than overwrite it, and that this is fast for large numbers of samples.

sbillinge:
@danielballan (and @tacaswell ?) please could you look at @chiahaoliu comment about why we needed `.get_md(1)`, and whether there is a cleverer way to do this that doesn't break the standard python interface?  If you can't think of anything please say so and we will pull this version into the current release.  Thx so much.

xpdacq/beamtime.py
```diff
@@ -244,6 +244,13 @@ def _clean_info(obj):
     """ stringtify and replace space"""
     return str(obj).strip().replace(' ', '_')
 
+class MDOrderedDict(OrderedDict):
```
CJ-Wright: You might want to call it something different. The issue being addressed here, that python has overloaded `[bla]` to both get index `bla` from list and get value associated with key `bla` and that the dictionary usage took precedence, is a bit more generic than simply for metadata. Would it be possible name this something along the lines of `SliceableOrderedDict` and have a `get_slice` method? I think this may head off some of the unique syntax issues (it may even be a nice thing for other developers to use).

CJ-Wright: Also does this have an independent test?

danielballan:
I think samples as a dict (generally, dict-like object) keyed on sample name is sensible. Two questions about what to do with position-based indexing:

1. Is position-based lookup needed at all? When it is desirable to look up a sample based its position in the list (or OrderedDcit) rather than its name? Is the order significant to the user?
2. The standard Python equivalent of your `d.get_md(i)` is `d[list(d)[i]]`. Is the position-based lookup common enough to be worth naming a new method to do this?

chiahaoliu:
Thanks @danielballan for a very insightful comment and let me extend the discussion by answering one of these questions:

1. Is position-based lookup needed at all? 
* the motivations of setting a special method for enable position-based lookup are:
  * we allow people to index based on position, i.e. ``xrun(1,5)``, so user might expect the same logic when they want to interrogate the metadata.
  * when we do ``print(bt)`` (or ``bt.list()``), we would have an output with index:
  ````
  Samples:
  0: activated_carbon_1
  1: activated_carbon_2
  ....
  ````
  * for current version, ``bt.samples`` is a list, so ``bt.samples[1]`` would return the sample object hence metadata stored in it.

* but as Dan pointed out, user can use standard key-pair lookup method for dict-like object.

Due to the speed issue described above, we might want to move to dict-like structure. Then the question I would like to ask is:
* is it fair to ask user copy and paste ``sample_name`` when they want to examine the metadata, like ``bt.samples['activated_carbon_1']`` ?

danielballan:
Point of info: also remember that tab -complete works here, as in
bt.samples['a<TAB> so the user might not even need to copy/paste.

chiahaoliu:
based on discussion, we might keep the special ``get_md`` method. It is more magic but also user friendly. Also standard python ``dict`` lookup still works ( thanks @danielballan for informative comments), user can choose whichever method he wants

xpdacq/beamtime.py
```diff
@@ -337,6 +338,12 @@ def _clean_info(obj):
...
+    def get_md(self, ind):
+        """ special method to get metadata of a objects based on index """
```
CJ-Wright: special method to get metadata of sample object based on bt.list index

sbillinge:
Simon was approver but using CJ computer, so don't need approval of CJ to pull this in... therefore merging

PR #321, opened 2016-11-04
STY: add landscape yml so we don't error on python 3 things

CJ-Wright:


CJ-Wright:
@sbillinge @sbanerjee23 @chiahaoliu Can I get a fast track on this? The hope is to rebase all the current PRs onto this so we can get more realistic landscape reports.

PR #322, opened 2016-11-07
Fix min expo time

### LINKED ISSUE #282: capture exception when creating ScanPlan with exposure time shorter than frame acquire time
throw an exception and ask user to check if frame acquire time has been set to correct value.

chiahaoliu:
closes #282 

This PR also includes the enhancement for simulation objects. Now they are closer to realistic experiment object at the syntax level. A more rigorous way of finding the best dark frame using current logic is also introduced.

@sbillinge  @CJ-Wright  code is ready for review. Let me know if this PR need to squash


chiahaoliu:
@sbillinge error message is updated and ready for review again

PR #323, opened 2016-11-18
Add temperature list scan

### LINKED ISSUE #268: add temperature list scan to one of xpdAcq predefined scans

chiahaoliu:
closes #268 

@sbillinge @CJ-Wright  ready for review

### DIFF
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -170,13 +170,14 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
 
 def T_list(dets, exposure, T_list):
     """
-    Scan over the temperature controller in a list of user defined steps
+    Taking images from the area detector while scaning over the temperature
+    controller in a list of user defined steps
 
     Parameters
     ----------
     dets : list
-        list of objects represent instrument devices. In xpdAcq, it is
-        default to area detector.
+        list of objects that represent instrument devices. In xpdAcq, it is
+        defaulted to area detector.
     exposure : float
         total time of exposure in seconds for area detector
     T_list : list
@@ -191,7 +192,7 @@ def T_list(dets, exposure, T_list):
         >>> glbl.area_det
         >>> glbl.temp_controller
 
-    To figure interrogate devices that are currently in use.
+    To interrogate which devices are currently in use.
     """
 
     pe1c, = dets
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -168,10 +168,11 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
     yield from plan
 
 
-def T_list(dets, exposure, T_list):
-    """
-    Taking images from the area detector while scaning over the temperature
-    controller in a list of user defined steps
+def Tlist(dets, exposure, T_list):
+    """defines a flexible scan with user-specified temperatures
+
+    A frame is exposed for the given exposure time at each of the
+    user-specified temperatures
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


xpdacq/beamtime.py
```diff
@@ -168,6 +168,65 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
     yield from plan
 
 
+def T_list(dets, exposure, T_list):
```
sbillinge: To keep things consistent, should we call it `Tlist`?  Since we use `Tramp` and not `T_ramp`.

xpdacq/beamtime.py
```diff
@@ -168,6 +168,65 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
...
+    """
+    Scan over the temperature controller in a list of user defined steps
```
sbillinge: write a brief description of the function on the top line (same line as the """) then blank line, then a longer description if needed, then the arguments and what is returned.

I am not sure what this does, but let me take a guess here:

```
"""defines a flexible scan with user-specified temperatures

A frame is exposed for the given exposure_time at each of the
temperatures in the T_list

...

```

xpdacq/beamtime.py
```diff
@@ -168,6 +168,65 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
...
+
+    Parameters
+    ----------
+    dets : list
+        list of objects represent instrument devices. In xpdAcq, it is
+        default to area detector.
+    exposure : float
+        total time of exposure in seconds for area detector
+    T_list : list
+        a list of temperature steps that you wish to scan over
```
sbillinge: is it a list of temperature steps, or a list of temperatures where a scan is to be run?

xpdacq/beamtime.py
```diff
@@ -168,6 +168,65 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
...
+
+    Note
+    ----
+    area detector and temperature controller that would be triggered will
+    always be the one configured in global state. Please using following
```
sbillinge:
```
always be the one configured in global state. 
To find out which these are, use the following commands:
"""

File: xpdacq/beamtime.py [hunk 5]
```diff
@@ -168,6 +168,65 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
...
+        >>> glbl.area_det
+        >>> glbl.temp_controller
+
+    To figure interrogate devices that are currently in use.
+    """
+
+    pe1c, = dets
+    # setting up area_detector and temp_controller
+    (num_frame, acq_time, computed_exposure) = _configure_pe1c(exposure)
+    motor = glbl.temp_controller
```
sbillinge: why on earth are you using `motor` for the variable name here.  There is no motor in this problem.
Better to call it `t_controller` or something.

xpdacq/beamtime.py
```diff
@@ -168,6 +168,66 @@ def Tramp(dets, exposure, Tstart, Tstop, Tstep, *, md=None):
...
+
+    To interrogate which devices are currently in use.
+    """
+
+    pe1c, = dets
+    # setting up area_detector and temp_controller
+    (num_frame, acq_time, computed_exposure) = _configure_pe1c(exposure)
+    motor = glbl.temp_controller
+    # manage md
+    bluesky_parent_md = {'detectors': [det.name for det in dets],
```
danielballan: All this is already provided by list_scan itself. Why copy/paste here? The only metadata you need to override is plan_name, which should match the function name (Tlist or T_list, whichever you go with).

PR #324, opened 2016-11-20
refactor glbl class

### LINKED ISSUE #292: reload glbl options after exiting out ipython

chiahaoliu:
closes #292 

Include customized ``YamlClass`` that is able to automatically save changes when attributes are updated and use this functionality to make reload of ``glbl`` class possible.

ready for review: @sbillinge @CJ-Wright

### DIFF
--- a/xpdacq/glbl.py
+++ b/xpdacq/glbl.py
@@ -80,7 +80,9 @@ _EXPORT_TAR_DIR = [CONFIG_BASE, USERSCRIPT_DIR]
 
 
 class Glbl(YamlClass):
-    """A class stores global options of xpdAcq"""
+    """
+    A class stores global options of xpdAcq
+    """
     _is_simulation = simulation
     beamline_host_name = BEAMLINE_HOST_NAME
     # directory names
@@ -135,13 +137,14 @@ class Glbl(YamlClass):
                  'tri_offset': 13, 'v_asym': 0,
                  'alpha': 2.5, 'tmsk': None}
 
-    def __init__(self, filepath, internal_dict={}):
+    def __init__(self, filepath, internal_dict=None):
+        if internal_dict is None:
+            internal_dict = {}
         super().__init__(internal_dict)
         self._filepath = filepath
 
     def allowed_attributes(self):
-        """define a list of attribute names that are allowed to be
-        yamlize"""
+        """list of attribute that are allowed to be serialized"""
         return ['auto_dark', 'dk_window', '_dark_dict_list',
                 'shutter_control', 'auto_load_calib',
                 'calib_config_name','calib_config_dict',
--- a/xpdacq/glbl.py
+++ b/xpdacq/glbl.py
@@ -1,7 +1,6 @@
 import os
 import socket
 import time
-import yaml
 from time import strftime, sleep
 
 from xpdacq.yamlclass import YamlClass
@@ -143,8 +142,8 @@ class Glbl(YamlClass):
         super().__init__(internal_dict)
         self._filepath = filepath
 
-    def allowed_attributes(self):
-        """list of attribute that are allowed to be serialized"""
+    def tracked_attributes(self):
+        """list of attribute being tracked and serialized"""
         return ['auto_dark', 'dk_window', '_dark_dict_list',
                 'shutter_control', 'auto_load_calib',
                 'calib_config_name','calib_config_dict',
--- a/xpdacq/glbl.py
+++ b/xpdacq/glbl.py
@@ -99,7 +99,11 @@ class Glbl(YamlClass):
     experiment_dir = EXPERIMENT_DIR
     scanplan_dir = SCANPLAN_DIR
     allfolders = ALL_FOLDERS
-    archive_dir = USER_BACKUP_DIR
+    if simulation:
+        archive_dir = os.path.join(BASE_DIR,
+                                   'userSimulationArchive')
+    else:
+        archive_dir = USER_BACKUP_DIR
     # on/off and attributes for functionality
     auto_dark = True
     dk_window = DARK_WINDOW
@@ -117,9 +121,6 @@ class Glbl(YamlClass):
     mask_name = MASK_NAME
     calib_config_dict = None
     mask = None
-
-    if simulation:
-        archive_dir = os.path.join(BASE_DIR, 'userSimulationArchive')
 
     # object should be handled by ipython profile
     db = None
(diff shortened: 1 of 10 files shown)


xpdacq/glbl.py
```diff
@@ -137,24 +137,35 @@ class Glbl:
...
-        self._frame_acq_time = frame_acq_time
+    def __init__(self, filepath, internal_dict=None):
+        if internal_dict is None:
+            internal_dict = {}
+        super().__init__(internal_dict)
+        self._filepath = filepath
+
+    def allowed_attributes(self):
+        """list of attribute that are allowed to be serialized"""
+        return ['auto_dark', 'dk_window', '_dark_dict_list',
```
CJ-Wright: If we are going to hard code this, can we put this as a class variable?

xpdacq/yamlclass.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# See AUTHORS.txt for a list of people who contributed.
+# See LICENSE.txt for license information.
+#
+##############################################################################
+
+import os
+import yaml
+
+class YamlClass:
```
CJ-Wright: what happens with the following?
```python
world = 1
a = YamlClass({'hello': world})
world = 2
```
Is the file updated?

xpdacq/yamlclass.py
```diff
@@ -0,0 +1,64 @@
...
+    special class automatically yamlize user-defined attributes
+    if they are updated
+    """
+
+    def __init__(self, internal_dict=None):
+        # dict stores values of valid attributes
+        if internal_dict is None:
+            internal_dict = {}
+        self._internal_dict = internal_dict
+        for key in self.allowed_attributes():
```
CJ-Wright: Why is `allowed_attributes` a method and not a attribute itself?

CJ-Wright: Can we change the name `allowed_attributes` to `tracked_attributes`?

xpdacq/yamlclass.py
CJ-Wright: Is this similar to https://traitlets.readthedocs.io/en/stable/api.html

xpdacq/yamlclass.py
```diff
@@ -0,0 +1,64 @@
...
+class YamlClass:
+    """
+    special class automatically yamlize user-defined attributes
+    if they are updated
+    """
+
+    def __init__(self, internal_dict=None):
+        # dict stores values of valid attributes
+        if internal_dict is None:
+            internal_dict = {}
```
CJ-Wright: Why not just default the `internal_dict` to `{}`?

xpdacq/glbl.py
```diff
@@ -137,24 +137,35 @@ class Glbl:
                  'tri_offset': 13, 'v_asym': 0,
                  'alpha': 2.5, 'tmsk': None}
 
-    def __init__(self, frame_acq_time=FRAME_ACQUIRE_TIME):
-        self._frame_acq_time = frame_acq_time
+    def __init__(self, filepath, internal_dict=None):
```
CJ-Wright: See below comment about init to `{}`

xpdacq/glbl.py
```diff
@@ -1,9 +1,11 @@
 import os
 import socket
 import time
-from unittest.mock import MagicMock
+import yaml
```
CJ-Wright: Do we need this import? Isn't the point of the YamlClass so we don't have to deal with yaml at this level?

CJ-Wright:
How does this interact with xpdAcq/xpdAn#30?

xpdacq/glbl.py
chiahaoliu: please see: 
http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

It is a convention to avoid confusion on default values.

xpdacq/yamlclass.py
chiahaoliu: please see: 
http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

It is a convention to avoid confusion on default values.

xpdacq/yamlclass.py
chiahaoliu: yes, sounds like a better name.

xpdacq/glbl.py
chiahaoliu: I want to prevent it from being updated. If it is a class variable, user could potentially update it (and break it)

xpdacq/yamlclass.py
chiahaoliu: please see the comment above. I want to prevent it from being updated.

xpdacq/yamlclass.py
chiahaoliu: No, it wouldn't be updated.

But why should we expect it to be updated? When you create a ``dict``, it is the value of input variable being passed, not the variable itself.

````
In [26]: word =1

In [27]: a = {'hello':word}

In [28]: a
Out[28]: {'hello': 1}
````

Is there any specific reason you want this logic to work? (``weakref``?)

chiahaoliu: can you elaborate a bit more? thanks.

chiahaoliu:
@CJ-Wright we are not at the stage of ``ChainMap`` yet, but basic implementation is ready.

xpdacq/glbl.py
CJ-Wright: Fair enough

xpdacq/yamlclass.py
CJ-Wright: I had just realized that traitlets might be able to do much of the things you were looking for this class to do. The basic idea would be that the `Glbl` class inherets from `HasTraits` then all the traits can be tracked using traitlets. Then use the `observe` functionality in traitlets to save the data down every time it changes.

CJ-Wright: Sorry, never mind this comment I forgot a bunch of things about how python objects work.

xpdacq/glbl.py
CJ-Wright: What if the user wanted to track a new variable?

xpdacq/yamlclass.py
chiahaoliu: That's true. Tracking functionality is definitely what we are looking for but this great power also comes with a price. We need to specifically define ``type``, allowed value and corresponding decorator for each of attribute. I would not go for this road at this stage (or at least not in this PR). ``Traits`` would be useful for future setting.

CJ-Wright: OK

sbillinge:
Please pull this in if you are ready to.  It looks good to me.

PR #327, opened 2016-12-15
metadata schema

### LINKED ISSUE #314: ENH: formalize required fields into schema
We should formalize what fields we need for each run start (PI name, sample name, saf#) in a json schema so that we can validate the results, also this will help with building test suites, as we can build fake MDS documents with the correct fields.

chiahaoliu:
closes #314 

add the schema for current metadata format

@sbillinge @CJ-Wright  PR ready for review


xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+      "type": "number",
+      "description": "x-ray wavelength"
+    },
+    "bt_safN":{
+      "type": "string",
+      "description": "Safety Approval Number of beamtime"
+    },
+    "bt_experimenters":{
+      "type": "array",
+      "description": "list of experimenter names in [first, last,...] format"
```
CJ-Wright: Can this be a list of tuples?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+    "mask":{
+      "type": "array",
+      "description": "sparse representation of the mask generated by auto-mask function"
+    },
+    "md_schema_version":{
+      "type": "string",
+      "description": "metadata version being used by xpdAcq"
+    }
+  }
+}
```
CJ-Wright: Is there a list of required keys like in [event-store](https://github.com/NSLS-II/event-model/blob/master/event_model/schemas/event.json#L32)?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+      "type": "array",
+      "description": "list of possible structure filenames"
+    },
+    "collaborators":{
+      "type": "array",
+      "description": "list of collaborator names"
+    },
+    "database_id":{
+      "type": "object",
+      "description": "dictionary of structure uid to different database"
```
CJ-Wright: This might be a bit tricky and might need to be a list of ids (if you have different samples in one run)

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+      "type": "array",
+      "description": "list of sample maker names "
+    },
+    "sa_uid":{
+      "type": "string",
+      "description": "auto-generated uid for Sample object"
+    },
+    "cif_name":{
+      "type": "array",
+      "description": "list of possible structure filenames"
```
CJ-Wright: We might want to have this in post acq workflow. If it is in the acq workflow it might need to be a list of lists. Also will filenames be enough? If we move the data around will the files move with the data? What if others have different filenames for the same file? If we want to capture this then we might think about using FileStore or something.

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+      "type": "array",
+      "description": "list of experimenter names in [first, last,...] format"
+    },
+    "bt_uid":{
+      "type": "string",
+      "description": "auto-generated uid for Beamtime object"
+    },
+    "sample_name":{
+      "type": "string",
+      "description": "name of sample being measured"
```
CJ-Wright: What about multiple samples?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+      "type": "object",
+      "description": "dictionary of structure uid to different database"
+    },
+    "lead_experimenter":{
+      "type": "array",
+      "description": "list of lead experimenter names"
+    },
+    "notes":{
+      "type": "string",
+      "description": "memorable note for sample"
```
CJ-Wright: Should this be "memorable note for experiment"? There may be more than one sample in the run.

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+    },
+    "sp_computed_exposure":{
+      "type": "number",
+      "description": "program computed exposure time"
+    },
+    "sp_type":{
+      "type": "string",
+      "description": "ScanPlan type of xpdacq"
+    },
+    "sp_startingT":{
```
CJ-Wright: This is fine, but I wouldn't go too far in specifying such scan plan specific things. It is fine for the users to inject their own metadata and it is ok if we don't validate that injected metadata.

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
+{
+  "type": "object",
+  "properties":{
+    "bt_piLast":{
+      "type": "string",
+      "description": "Last name of PI"
+    },
+    "bt_wavelength":{
```
CJ-Wright: I assume that all these `"bt_something"` fields are supposed to be defined on a per beamtime level. Is wavelength on a per beamtime level?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+    },
+    "collaborators":{
+      "type": "array",
+      "description": "list of collaborator names"
+    },
+    "database_id":{
+      "type": "object",
+      "description": "dictionary of structure uid to different database"
+    },
+    "lead_experimenter":{
```
CJ-Wright: shouldn't this be under `bt_`?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
+{
+  "type": "object",
+  "properties":{
+    "bt_piLast":{
```
CJ-Wright: (I may regret this) Would it be possible to put all these beamtime level things in their own dictionary, inside the header? This way when people ask for information about their beamtime we can just hand them the contents on their beamtime dictionary. The potential for regret stems from searching in nested dicts requires mongo syntax (I think) which the users are most likely less interested in (especially as I am pushing `bt_piLast' as an important filter for all the fuzzy searching capacity).

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
+{
```
CJ-Wright: Can we also add a facility and beamline key? (if they aren't there already)

xpdacq/md_schema_v1.json
chiahaoliu: It used to be a tuple, but we found that would introduce search difficulty. By making name as a list, mongo allows ``in`` logic by default.

xpdacq/md_schema_v1.json
chiahaoliu: currently, we haven't supported for multiple samples per run (or at least not explicitly). If we have robot/sample container running which allows multiple changing samples without human labor, I am happy to make it an ``array``.

xpdacq/md_schema_v1.json
chiahaoliu: same reason as before

xpdacq/md_schema_v1.json
CJ-Wright: Ugh, that is annoying (curse you mongo gods). [It feels weird](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) but I guess we don't have better options. @tacaswell thoughts?

xpdacq/md_schema_v1.json
CJ-Wright: Didn't Anton run an experiment with multiple "samples"?

xpdacq/md_schema_v1.json
chiahaoliu: what if the user wants to search his Tramp started from 75K? and actually this is injected by ScanPlan.

xpdacq/md_schema_v1.json
CJ-Wright: If it is in their spreadsheet why are we tracking it? Can we just provide a uid to their spreadsheet or vice versa and have the user handle it? (they already need to do it with the files) I agree that long term the non-file based workflow should fix this (although we need to think about non-filebased sample management/tracking, maybe amostra?)

>

xpdacq/md_schema_v1.json
chiahaoliu: all the ``bt_sth`` are the metadata fields stored under ``Beamtime`` object and all of the fields will be dumped into ``mds`` per run.

wavelength might not be unique per beamtime and we actually allow it to be updated dynamically if needed

xpdacq/md_schema_v1.json
CJ-Wright: That is fine, we are not restricting the search to only these keys, we are just specifying what keys we can expect to be there. The user will still be able to search based off of whatever keys are in the metadata, if they put it there or not. There is nothing stopping us from validating this when we expect it (one could make the point that every scan which has specialty data should have its own validation).

xpdacq/md_schema_v1.json
chiahaoliu: we have ``bt_experimenters`` which is a list of names and it indicates people would participle the beamtime. this field records the person who is mainly responsible fo certain sample.(again, it reads from spreadsheet)

xpdacq/md_schema_v1.json
CJ-Wright: Sorry I didn't understand the short hand (maybe put a comment in the schema that all things with `bt_` are managed by the beamtime object?)

xpdacq/md_schema_v1.json
chiahaoliu: yup, you got the point. we had nested ``dict`` at first but it turns out searching nesting ``dict`` was a huge pain.

We can discuss about it again since now db has filter method

CJ-Wright: Hmm, ok we should definitely discuss it after my exams if you have time.

xpdacq/md_schema_v1.json
chiahaoliu: yes, it is injected for every scan at XPD so I didn't list it. I can add it if it's helpful

CJ-Wright: The facility is injected as well? Huh, didn't know that. Formally we should validate it but it should be ok if we don't.

xpdacq/md_schema_v1.json
CJ-Wright: Hmm, [this doesn't look too bad](https://docs.mongodb.com/manual/core/document/#dot-notation) if `db()` supports it.

chiahaoliu:
In general, I am not sure if we are ready to use validate method as this stage as you can see based on the review comments about, we might need more iterations on how to format metadata. But at least we have a summarized table for all fields could appear.

Also we tried nested dictionary before, but it caused difficulties in searching. Now we have different tool, so we should discuss more on that.

xpdacq/md_schema_v1.json
CJ-Wright: Another way to handle this might be to have a map which extracts the beamtime specific keys from the flat dictionaries.

xpdacq/md_schema_v1.json
tacaswell: For these sorts of things do not rely on the implementation to always provide it, specify that it should be there and check.

"trust but verify"

xpdacq/md_schema_v1.json
tacaswell: See https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

I suggest just storing as `[name1, name2, name3, ...]`, this schema should not be in the business of encoding the semantics of names (I go by 'Thomas A Caswell' or 'Thomas Caswell' or 'Thomas Alexander Caswell', all of which have different semantic than 'Stefan van der Walt' (lead skimage dev)).

xpdacq/md_schema_v1.json
tacaswell: wrong meaning of the word 'run', each sample would get it's own header + run uid + ...

tacaswell:
Just a reminder that the run engine has a hook for exactly this sort of validation.

xpdacq/md_schema_v1.json
CJ-Wright: @tacaswell I just linked to that in the "It feels weird".

xpdacq/md_schema_v1.json
CJ-Wright: What do you mean? Can't we have one scan which takes data on multiple samples?

xpdacq/md_schema_v1.json
tacaswell: ha!

xpdacq/md_schema_v1.json
tacaswell: You _can_, but you probably shouldn't for these meta-data reasons.  There is already a layer that provides nice iteration over / selecting out of (which is by header) so put really obvious distinctions at that that level.

headers should be granular enough that you do not do a whole lot of event-level filtering and coarse enough that you are not always merging headers in the first round or two of analysis.

ex doing a temperature ramps over a bunch of samples, doing one run per T ramp may make sense if once the sample is mounted and good probably going to stay good and the goal is to quickly get to some temperature v devrived_value curve and over the course of a beam time you will take many many such curves.  On the other hand, at any temperature the sample may explode or degrade, etc and over the course of a single beamtime you will be taking many readings at a single temperature and then at the end assembling / filtering all of them into a single T v derived_value curve, then one run per temperature point makes more sense.

xpdacq/md_schema_v1.json
tacaswell: Because a spread sheet is a great data-entry tool, but not a great data-preservation tool.  You want to capture the value of the meta-data at the time of data-collection, not what they changed it to later.

xpdacq/md_schema_v1.json
CJ-Wright: How would you handle an array of samples? Eg I have a square and it has 100 samples on it arranged in a grid. (Currently we do this with grid scan I think)

CJ-Wright: Is there a good way to get 100 runs like that?

CJ-Wright: Awesome! Thank you!

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+    },
+    "sc_dk_field_uid":{
+      "type": "string",
+      "description": "uid of dark frame being associated by xpdAcq"
+    },
+    "calibration_md":{
+      "type": "object",
+      "description": "dictionary of calibration parameters linked by xpdAcq"
+    },
+    "calibration_collection_uid":{
```
CJ-Wright: Can we also have something along the lines of 'is_detector_calibration'?

xpdacq/md_schema_v1.json
CJ-Wright: What about the the inverse? A temperature ramp on the outside and a outer product on the inside?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+    },
+    "sp_computed_Tstep":{
+      "type": "number",
+      "description": "program computed temperature step of xpdAcq Tramp ScanPlan"
+    },
+    "sp_Nsteps":{
+      "type": "number",
+      "description": "program computed number of steps of xpdAcq Tramp ScanPlan"
+    },
+    "dark_frame":{
```
CJ-Wright: change to `is_dark`?

xpdacq/md_schema_v1.json
```diff
@@ -0,0 +1,153 @@
...
+      "type": "boolean",
+      "description": "tag to label dark frame"
+    },
+    "is_background":{
+      "type": "boolean",
+      "description": "tag to label background scan"
+    },
+    "sc_dk_field_uid":{
```
CJ-Wright: change to `dark_collection_uid`?

xpdacq/md_schema_v1.json
CJ-Wright: change to `detector_calibration_collection_uid`

CJ-Wright: Can we also have `energy_calibration_collection_uid` and `is_energy_calibration`?

CJ-Wright:
This most likely needs to exist as its own independent repository.

CJ-Wright:
This now has its own repo! see xpdAcq/xpdSchema#1

PR #339, opened 2016-12-21
Fix md parser

### LINKED ISSUE #283: revisit behavior of import_sample_info
mongo doesn't allow `.` in the key field so we can not  have metadata like:

`{'sample_composition':{'0.8mm_glass_capillary': 1}}`

need to rethink the behavior carefully.

chiahaoliu:
closes #283 

as per discussion, add a helper function to recursively replace ``.`` with ``,`` in key fields to avoid error when interacting with ``mongo``

### DIFF
--- a/xpdacq/tests/test_tools.py
+++ b/xpdacq/tests/test_tools.py
@@ -1,5 +1,5 @@
 import pytest
-from xpdacq.tools import clean_dict
+from xpdacq.tools import regularize_dict_key
 
 @pytest.mark.parametrize("input_dict, output_dict",
                          [({'a': 1, 'b': 2, 'c': 3},
@@ -11,6 +11,6 @@ from xpdacq.tools import clean_dict
                           ({'a.foo': {'b.bar': '2.foo'}, 'c': '3.bar'},
                            {'a,foo': {'b,bar': '2.foo'}, 'c': '3.bar'})
                          ])
-def test_clean_dict(input_dict, output_dict):
-    clean_dict(input_dict, '.', ',')  # modify dict in place
+def test_regularize_dict_key(input_dict, output_dict):
+    regularize_dict_key(input_dict, '.', ',')  # modify dict in place
     assert input_dict == output_dict  # test equality
--- a/xpdacq/tests/test_tools.py
+++ b/xpdacq/tests/test_tools.py
@@ -1,19 +1,17 @@
+import pytest
+import copy
 from xpdacq.tools import clean_dict
 
-def test_clean_dict():
-    # case 1: oridinary dict, no changes
-    test = {'a': 1, 'b': 2, 'c': 3}
-    clean_dict(test, '.', ',')
-    assert test == {'a': 1, 'b': 2, 'c': 3}
-    # case 2: flat dict with target character
-    test = {'a.foo': 1, 'b.bar': 2, 'c': 3}
-    clean_dict(test, '.', ',')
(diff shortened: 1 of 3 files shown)


xpdacq/tools.py
```diff
@@ -0,0 +1,36 @@
...
+    Parameters
+    ----------
+    input_dict : dict
+        a dictionary going to be cleaned. it could be nested
+    target_chr : str
+        character that will be placed
+    replace_chr : str
+        character that is going to replace target character
+    """
+    # for safety, copy a new one
```
CJ-Wright: I'm confused, are you creating a new copy or mutating the current dict?

xpdacq/tests/test_tools.py
```diff
@@ -0,0 +1,19 @@
+from xpdacq.tools import clean_dict
+
+def test_clean_dict():
```
CJ-Wright: Can you set this up as a parameterized test? http://doc.pytest.org/en/latest/parametrize.html

xpdacq/tools.py
chiahaoliu: Sorry, that was the outdated comment, my mistake. This function will mutate current dict

chiahaoliu:
@CJ-Wright  test has been rewritten, please take a look if you have time, thanks.

xpdacq/tools.py
```diff
@@ -0,0 +1,34 @@
...
+##############################################################################
+def clean_dict(input_dict, target_chr, replace_chr):
+    """recursively replace target character in keys with desired one
+
+    Parameters
+    ----------
+    input_dict : dict
+        a dictionary going to be cleaned. it could be nested
+    target_chr : str
+        character that will be placed
```
sbillinge: placed -> replaced

xpdacq/tests/test_tools.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+import copy
+from xpdacq.tools import clean_dict
+
+@pytest.mark.parametrize("input_dict, output_dict",
+                         [({'a': 1, 'b': 2, 'c': 3},
+                           {'a': 1, 'b': 2,'c':3}),
```
CJ-Wright: Space between `:` operator

xpdacq/tools.py
```diff
@@ -0,0 +1,35 @@
...
+#                        New York.
+#                   All rights reserved
+#
+# File coded by:    Timothy Liu
+#
+# See AUTHORS.txt for a list of people who contributed.
+# See LICENSE.txt for license information.
+#
+##############################################################################
+def clean_dict(input_dict, target_chr, replace_chr):
```
CJ-Wright: Would it be possible to name this something a bit more descriptive? Like `replace_dict_keys`? Cleaning a dictionary can mean a lot of things.

chiahaoliu: actually this is a good point

tacaswell:
Why chose `,` as the replacement character?   This means you can not safely reverse the cleaning.  For a while we had code to do this but replaced it with `__PLACEHOLDER__` or something like that.

We ended up deciding that this sort of cleaning was more trouble than it was worth (as making user data round-trip required pretty annoying munging on both in and out).  As this is implemented user data _will not_ round trip and users will be confused.  We decided the best path forward here was to just expose the mongo limitation (on `.` in keys (which is not that unreasonable as it provides a nice way to get at nested values)) all the way up to the users.

Strongly suggest that you do the same.

PR #343, opened 2016-12-27
ENH: FileStore integrated reader

chiahaoliu:
This short PR is not in the original release scope but I found this change could a complete cycle possible, from generating images, pushing to local ``db`` and querying them. This can potentially make our tests a lot more robust.

@sbillinge @CJ-Wright PR is ready for review

### DIFF
--- a/xpdacq/simulation.py
+++ b/xpdacq/simulation.py
@@ -52,29 +52,19 @@ class SimulatedCam:
 
 
 # define simulated PE1C
-class SimulatedPE1C(be.Reader):
+class SimulatedPE1C(be.ReaderWithFileStore):
     """Subclass the bluesky plain detector examples ('Reader');
 
     also add realistic attributes.
     """
 
-    def __init__(self, name, read_fields):
+    def __init__(self, name, read_fields, fs):
         self.images_per_set = PutGet()
         self.number_of_sets = PutGet()
         self.cam = SimulatedCam()
         self._staged = False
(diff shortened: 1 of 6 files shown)


xpdacq/simulation.py
```diff
@@ -190,6 +180,7 @@ def insert_imgs(mds, fs, n, shape, save_dir=tempfile.mkdtemp()):
 
 # instantiate simulation objects
 db = build_pymongo_backed_broker()
-pe1c = SimulatedPE1C('pe1c', {'pe1_image': lambda: 5})
+db.fs.register_handler('RWFS_NPY', be.ReaderWithFSHandler)
+pe1c = SimulatedPE1C('pe1c', {'pe1_image': lambda: np.ones((5,5))}, fs=db.fs)
```
CJ-Wright: If you want you can steal the image cycle logic from [xpdSim](https://github.com/xpdAcq/xpdSim/pull/13/files#diff-a2f91388ba37eebfc35980ace3307f21R15) and the use of a data directory to use real images.

chiahaoliu: There is the plan for fully integrated with ``xpdSim`` (issue #352) as we discussed a while ago. But I would prefer to do it in separate PR to minimize potential effects.

chiahaoliu:
@CJ-Wright do you have any comment on this PR except for the image logic?

CJ-Wright:
Nope it looks good. Merging.

PR #344, opened 2016-12-28
Revert "Fix md parser"

CJ-Wright:
Reverts xpdAcq/xpdAcq#339
Since @tacaswell had objections to the md parser fix I thought that I should open a revert PR just in case.


chiahaoliu:
repost @tacaswell comment to here:

""""
Why chose , as the replacement character? This means you can not safely reverse the cleaning. For a while we had code to do this but replaced it with __PLACEHOLDER__ or something like that.

We ended up deciding that this sort of cleaning was more trouble than it was worth (as making user data round-trip required pretty annoying munging on both in and out). As this is implemented user data will not round trip and users will be confused. We decided the best path forward here was to just expose the mongo limitation (on . in keys (which is not that unreasonable as it provides a nice way to get at nested values)) all the way up to the users.

Strongly suggest that you do the same.
""""

* use "," to replace "." is based on how to make this cleaning less obvious. We borrowed this idea from how European people write :p. Maybe it is less rigorous from programming aspect.

* from the user interface point of view, xpdAcq user would only face this problem when creating a ``Sample`` object. A preferred workflow (maybe 90% of the time) is to import and parse metadata from a spreadsheet. Therefore, user would still have his original information stored in that spreadsheet, cleaning only happens to dictionary stored inside ``Sample`` object. Of course, I can't completely rule out the possibility of user  doing ``sample_md={blah...}, Sample(bt, sample_md)``

* expose mongo limitation also makes sense too. If it is that case, we might need to rethink the spreadsheet template.

CJ-Wright:
At the very least can we get some additional tests with a fake databroker inserting these modified dicts and retrieving them?

I am a bit worried that the search replacement being discussed may be quite tricky to pull off. Eg what happens if I have both 0.8 (which gets converted) and 0,8 as keys?
What happens if I have actual commas in the the key I am looking for, do they get converted into periods by mistake?

chiahaoliu:
I thought the point of replacement is we can not have any comma in the key field as Mongo reserves it as an operator to unpack nested dict?

ref: https://github.com/NSLS-II/event-model/pull/8

For the test, we can do in the ``xpdAn`` side

CJ-Wright:
I thought that the problem was periods not commas.

My point was that with the mapping of periods to commas you run the risk of creating non-unique keys.
For example what if my dict looks like this `{0.8: 'foo', '0,8': 'bar'}? Can the conversion be done? If so how do I address 'foo' and not 'bar'?

chiahaoliu:
yeah, that problem is period ``.``, not comma ``,``. sorry typo there.

regarding the point, I think the sequence should be:
``{0.8: 'foo', '0,8': 'bar'}`` --> nothing happen as ``replace`` can't work on non-str (which is not captured in current code, I will issue a PR to fix this. Thanks for reminding me)
``{'0.8': 'foo', '0,8': 'bar'}`` --> ``event_model`` will raise when you send to RE, scan won't even run.

CJ-Wright:
Sorry, I put in a typo I meant the second one. I thought this code was supposed to address the second case and convert the `.` to a `,` so it is RE safe.

chiahaoliu:
Here is the output:

````
In [1]: from xpdacq.tools import regularize_dict_key

In [2]: aa = {'0.8':'foo', '0,8':'bar'}

In [3]: regularize_dict_key(aa, '.', ',')
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-3-fb22c636aeb1> in <module>()
----> 1 regularize_dict_key(aa, '.', ',')

/home/timothy/Documents/Repo/xpdAcq/xpdacq/tools.py in regularize_dict_key(input_dict, target_chr, replace_chr)
     26         character that is going to replace target character
     27     """
---> 28     for k, v in input_dict.items():
     29         if isinstance(v, dict):
     30             if isinstance(k, str):

RuntimeError: dictionary changed size during iteration
````

````
In [5]: bb = {0.8: 'foo', '0.8': 'bar'}

In [7]: regularize_dict_key(bb, '.', ',')

In [8]: bb
Out[8]: {0.8: 'foo', '0,8': 'bar'}
````

Should we be happy that python capture for us? :p

CJ-Wright:
As a potential solution to the issues raised by @sbillinge can we provide the users with a spreadsheet checker well in advance? This way we can have the software yell at the users well before they even step foot on the experimental floor and let them decide how to fix things. This way we don't have to write too much new code (just a validator that everything has no periods), the translation between spreadsheet to Mongo is direct and unique, and the users get to decide what the appropriate way to handle their data is (rather than us just imposing a fairly arbitrary name mapping). Since the code should be fairly simple we could send it as a windows/mac/linux friendly executable which takes in as args the path/to/spreadsheet and gives an ok or tells you where things are problematic.

chiahaoliu:
All fair, so combining the comments above, I am proposing following logic:
1. have a standalone function called something like ``validate_spread_sheet_md``. It provides users a way to validate the metadata filled in a spreadsheet and it raises an ``Exception`` when it sees invalid metadata.(program stops every time it sees invalid md)
2. at the run time, every metadata dictionary goes into ``Sample`` object will be *mutated* automatically in order to keep the experiment going but program yields ``Warning`` to inform user (program goes but issue warning)

does it sound reasonable?

CJ-Wright:
That sounds ok to me, although as a user I would prefer that `validate_spread_sheet_md` collates the problems and then raises that way I can deal with all of them at once.

tacaswell:
@sbillinge This restriction is only on _keys_ not values.  What sort of things are people using as keys that must have a period in them?

This might also be a place to provide a nudge improve the metadata layout!

sbillinge:
@chiahaoliu @tacaswell if the dot restriction is only on _keys_ not _values_ then we are all ok.  I don't think we have any dots in any of our keys, and if we do we can easily do away with them, unless I am missing something.....

chiahaoliu:
we would only have problem in one place: kapton background (``Sample``), and I just realized there is a nicer way to change the fundamental parsing logic and solve all the issues. I am working on a PR to capture all these fixes, along with the suggested validating logic mentioned above.

CJ-Wright:
@chiahaoliu Does that mean I should merge the revert?

chiahaoliu:
@CJ-Wright yes, please. Though new PR will overlap with reverted a little bit.

sbillinge:
I am still worried that we have the wrong schema here. That the background
name is a key. Is that necessary?

CJ-Wright:
@sbillinge Can you point to place in the code where this happens? (potentially in the new PR) I don't know if I have enough context.

PR #349, opened 2016-12-28
FIX: only reload glbl if beamtime is already started

chiahaoliu:
That is a small bug I found during testing with local simulation. It causes an error if we ``icollection`` in but without already starting a beamtime. 

This PR is ready for review


sbillinge:
OK, pulling in.

CJ-Wright:
@sbillinge Did this get merged or just closed?

chiahaoliu:
This PR might be re accidentally closed instead of being merged. Reopen and put to review.

Issue #350, opened 2016-12-29
update example sample spreadsheet in light of new update form of sample import


sbillinge:
we should also revisit what fields we want in the schema

PR #351, opened 2016-12-30
FIX_md_parser: v2 

### LINKED ISSUE #283: revisit behavior of import_sample_info

chiahaoliu:
closes #283, #350 

This PR summarize suggested logics mentioned in the close PR #344.
It includes fixes on two parts:
1. fundamental fix for md_parser logic so that we won't encounter the same issue currently mentioned in issue #283. Therefore our example sheet doesn't need to be updated.
2. add validation option to ``import_sample_info`` function. User can use it to inspect metadata entered inside the spreadsheet.
3. add mutation logic when instantiating ``Sample`` so if user is not aware of invalid ``.`` in key fields, program shows warning(but not error) and mutates the metadata automatically to make experiments running.

This PR is ready for review, thanks.


chiahaoliu:
unknown error appears. close it for now and work on it tomorrow

chiahaoliu:
error comes from git confusion happened in rebase. Solved and reopen PR

xpdacq/utils.py
```diff
@@ -372,18 +360,26 @@ def create_yaml(self, bt):
...
-                         for el in self.parsed_sa_md_list if
-                         el['sample_name'].startswith('bkgd')]
-        for bkg_name in set(bkg_name_list):
-            bkg_dict = {'sample_name': bkg_name,
-                        'sample_composition': {bkg_name: 1},
-                        'is_background': True}
-            Sample(bt, bkg_dict)
+        for md_dict in self.parsed_sa_md_list:
+            sample_name = md_dict.get('sample_name')
+            if sample_name.startswith('bkgd'):
```
CJ-Wright: Is there a better way to be more explicit about this? (eg, have a field in the spreadsheet stating that a sample is a background?)

chiahaoliu:
close based on the discussion with @CJ-Wright  and @sbillinge. Replace with new PR

PR #359, opened 2017-01-10
Fix spreadsheet parser

### LINKED ISSUE #350: update example sample spreadsheet in light of new update form of sample import

chiahaoliu:
closes #350, #283 

This is the overall fix of importing metadata from the spreadsheet. This PR covers:

1. new logic of associating sample and background. It checks if ``bkgd_sample_name`` already appears as one of the ``sample_name`` while importing. If not, it issues a warning. It also checks if user are using mapping for high-dimensional sample metadata.

2. logic of replacing ``.`` with ``,`` while instantiating ``Sample`` objects. We wouldn't have this problem anymore but still add this functionality here for our wild users.

3. small enhancements:
   * adding our example spreadsheet under version control
   * make parsed string aligned with ``pdfgetx3``

PR is ready for review


sbillinge:
@chiahaoliu these review comments I typed in the office yesterday, but I never submitted the review.  I just submitted them.  I am not sure if they are already fixed or not, but please pull this in yourself if and when they are fixed.

chiahaoliu:
yes, these were fixed when we sit together in the office

PR #360, opened 2017-01-10
FIX: now we only load local things if we are running a simulation

CJ-Wright:


CJ-Wright:
I think this has been fixed elsewhere

chiahaoliu:
yes, thanks for closing

PR #377, opened 2017-03-03
Sim

CJ-Wright:
This needs some love. For some reason none of the files are being found in the install of xpdSim
@tacaswell @danielballan any ideas?


danielballan:
I see the problem: if you want the tests to be importable, you need to put an `__init__.py` in the directory. You also might need to add the package to setup.py, depending on how you set up on your setup.py.

PR #378, opened 2017-03-10
TST/CI: cover tests

CJ-Wright:


chiahaoliu:
conflict in py3.6, guess it's trivial but does it imply we need to review our package dependency?

CJ-Wright:
I think the issue is that a lot of the packages aren't currently built for 3.6? I know that pyFAI isn't (not that it is a dep for this, just an example) and the databroker stack might not be. I can check up on this, if this is the case I will try to repackage what I can and either a) set 3.6 to allow fail or b) remove 3.6

PR #390, opened 2017-03-28
v0.5.2

chiahaoliu:
this is the version currently used at the beamline. It's the derivative of current master, with bug-fix when working at the beamline and it includes the prototype of ``statTramp`` as well.

Originally it was pushed to upstream and I rebased in my fork. Hope this PR can be reviewed, then I can make a release tag along with greatly refined documentation. Thanks.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -49,7 +49,13 @@ def _update_dark_dict_list(name, doc):
     dark_dict = {'acq_time': acq_time, 'exposure': light_cnt_time,
                  'timestamp': doc['time'], 'uid': doc['run_start']}
     dark_dict_list.append(dark_dict)
-    glbl['_dark_dict_list'] = dark_dict_list  # update glbl._dark_dict_list
+    if doc['exit_status'] == 'success':
+        print('dark frame complete, update dark dict')
+        dark_dict_list.append(dark_dict)
+        glbl['_dark_dict_list'] = dark_dict_list  # update glbl._dark_dict_list
+    else:
+         print("INFO: dark scan was not successfully executed.\n"
+               "gobal dark frame information will not be updated!")
 
 
 def take_dark():
@@ -150,8 +156,10 @@ def _validate_dark(expire_time=None):
                                         time_diff))
     if qualified_dark_list:
         # sort wrt expo_diff and time_diff for best candidate
+        #best_dark = sorted(qualified_dark_list,
+        #                   key=lambda x: x[1] and x[2])[0]
         best_dark = sorted(qualified_dark_list,
-                           key=lambda x: x[1] and x[2])[0]
+                           key=lambda x: x[2])[0]
         best_dark_uid = best_dark[0]
         return best_dark_uid
     else:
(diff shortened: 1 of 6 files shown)


xpdacq/xpdacq.py
```diff
@@ -49,7 +49,13 @@ def _update_dark_dict_list(name, doc):
     dark_dict = {'acq_time': acq_time, 'exposure': light_cnt_time,
                  'timestamp': doc['time'], 'uid': doc['run_start']}
     dark_dict_list.append(dark_dict)
-    glbl['_dark_dict_list'] = dark_dict_list  # update glbl._dark_dict_list
+    if doc['exit_status'] == 'success':
+        print('dark frame complete, update dark dict')
+        dark_dict_list.append(dark_dict)
+        glbl['_dark_dict_list'] = dark_dict_list  # update glbl._dark_dict_list
+    else:
+         print("INFO: dark scan was not successfully executed.\n"
```
sbillinge: In general I want our docstrings to be somewhat user-facing/user-helping, especially error messages.  There is nothing more frustrating for a user to find out that something went wrong, with no hint about whether that matters and what to do next.

First, I would say that a failed dark warrants being a WARNING than INFO, doesn't it?  If it failed there must be something wrong.  If anything the print statement from the successful dark should be an INFO:   I also worry about chatter from the software when things are going smoothly....if the software is always outputting statements reassuring the user that things are going well, they won't read them, and then they will miss statements when things didn't go properly.  

Ideally there would be a way that statements went to a log file that is saved somewhere and could be reviewed when desired, but suppressed from being seen by the user under normal circumstances.

Then when there is a warning, please always think....if I were the user, what should I do if I see this warning "dark scan was not successfully executed."?

xpdacq/xpdacq.py
```diff
@@ -150,8 +156,10 @@ def _validate_dark(expire_time=None):
                                         time_diff))
     if qualified_dark_list:
         # sort wrt expo_diff and time_diff for best candidate
+        #best_dark = sorted(qualified_dark_list,
+        #                   key=lambda x: x[1] and x[2])[0]
         best_dark = sorted(qualified_dark_list,
-                           key=lambda x: x[1] and x[2])[0]
+                           key=lambda x: x[2])[0]
```
sbillinge: this seems to have an unbalance paren.  is that correct?

xpdacq/xpdacq.py
CJ-Wright: There is a logging module we could use for this.

CJ-Wright: There is a logging module we could use for this.

chiahaoliu: Unfinished header usually happens when the user aborts scan intentionally as suspending would not close the header. To increase the visibility of this error, I am thinking the official python ``warning`` module paired with ``logging``, so that user is aware of error and know where to look at the error logs.

xpdacq/xpdacq.py
danielballan: The matching paren is in line 161 after `sorted`.

danielballan:
> this is the version currently used at the beamline. 

@sbillinge Since we are deploying this at XPD, we want this to be merged and tagged. We can iterate from here.

sbillinge:
Yes, please go ahead (or probably past tense at this point), but to be clear, I also want it merged and tagged!

PR #396, opened 2017-06-23
 ENH: shutter plan

chiahaoliu:
Main feature of this PR is customized shutter control to ``Tlist`` and ``Tramp`` scan plan so that shutter is closed between each temperature points.

This PR also includes some housekeeping works regarding cleaning g commented lines, PEP8 style and tweaks to make tests more rigorous.

PR #397, opened 2017-06-25
CI: move around deps

CJ-Wright:


CJ-Wright:
The tests that fail are not the fault of the CI any longer, so I think that this is ready to merge.
@chiahaoliu @sbillinge thoughts?

chiahaoliu:
Agree. the fails comes are due to some changes in dependent packages and are addressed in PR #396

CJ-Wright:
Awesome, merging

Issue #399, opened 2017-07-18
ENH: improve end_beamtime


chiahaoliu:
As reported, ``_end_beamtime`` might take a long time if there were large amounts of images collected. If the previous user didn't run ``end_beamtime`` ahead of time, that might cause a delay for the next user. 

Current approach: 
    Bundle (but not compress) ``xpdUser`` to remote backup place
Bottleneck:
    The network speed between local workstation at XPD and the remote storage managed by NSLS-II.

Proposed solution:
    1. Since new beamtime generally starts with a fresh terminal and new user usually has less interest in the old beamtime, we can use the fact and instruct the user to run ``end_beamtime`` at a different terminal.  Therefore, this ``end_beatime`` can take indefinitely long time and the system dialogue about archiving wouldn't bother the new user.
    2. Exclude tiffs while archiving to avoid network issue and provide a tool for beamline staff to recover all the tiffs collected within certain beamtime.

Add-on:
   Progress bar. This can assure the user that program is still working.
   ref: https://pypi.python.org/pypi/progressbar2/3.20.0

CJ-Wright:
this might also work for a progress bar
https://pypi.python.org/pypi/tqdm

CJ-Wright:
Maybe we can just do a subprocess call to `rsync`?

PR #400, opened 2017-07-28
Enh endbeamtime v2

chiahaoliu:
This is a PR for issue #399. It implements the workflow of excluding tiffs during archiving. Another approach in the issue #399, leaving another terminal running the end_beamtime, is dangerous as next user might close it anyway and leave the remote backup process incomplete.

This PR also uses ``rsync`` to have nature progress report so that user can know the program is still running. Currently traivs build won't pass as required fixes on the dependencies reside in PR #396. Will rebase after it's merged in.

@sbillinge @CJ-Wright any thought?
also @CJ-Wright we might need to have a simple script for beamline staff to recover all the tiffs if we are going with this end_beamtime approach.

### DIFF
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -193,11 +193,12 @@ def _end_beamtime(base_dir=None, archive_dir=None, bto=None, usr_confirm='y'):
     """ funciton to end a beamtime.
 
     Detail steps are:
-        2) Arhcive ``xpdUser`` directory to remove backup
+        2) Archive ``xpdUser`` directory to remove backup
         3) Ask for user confirmation
         4.1) if user confirms, flush all sub-directories under
         ``xpdUser`` for a new beamtime.
-        4.2) if user doesn't confirm, leave ``xpdUser`` untouched.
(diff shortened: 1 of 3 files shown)


CJ-Wright:
What kind of information will we have to reconstruct the image files? Can I know all the header uids?

chiahaoliu:
I would expect user to provide his ``saf``, ``bt_piLast`` and ``date`` so that at our side, we should, equivalently, have all the headers.

sbillinge:
Before working on this, perhaps we should ask the instrument scientists if
they want it.   This was not high on my priority stack as I hadn't received
any complaints from the instrument scientists.  Have you received
requests/complaints from them?

Having said that, is the basic idea that instead of copying tiffs you copy
enough information that if the user asks for the tiffs they can be pulled
from db[] ?

In that case we should think of a few requests that users might make to
instrument scientists who have to restore data to users.   I am not sure
whether all tiffs that users save were created by save_tiff() logic (which
we should be able to recreate) and how many have been modified in fit2D and
so on (which we almost certainly can't).

S

chiahaoliu:
Got it. Set it to enhancement tag for now.

PR #408, opened 2017-08-22
FIX/ENH: facility md

### LINKED ISSUE #318: ENH: insert facility into header MD

### LINKED ISSUE #369: MTN: xpdSchema
We need to use and validate our run starts against the xpdSchema
We also need to keep track of the data required by the xpdSchema

chiahaoliu:
closes #318 also finish all the md requirements, hence closes #369


chiahaoliu:
@CJ-Wright ready for review

chiahaoliu:
@CJ-Wright finally fix ``xpdSim`` package resource. Up beamline md and ready for next round of review.

.travis.yml
```diff
@@ -25,7 +25,7 @@ addons:
 
 matrix:
   include:
-    - python: 3.5
+    - python: 3.6
```
CJ-Wright: Is the beamline py36? Was there a specific reason for this change? (not that I have a problem with it)

xpdacq/tests/test_xrun.py
```diff
@@ -48,6 +48,8 @@ def setUp(self):
         shutil.copyfile(src, os.path.join(glbl['import_dir'], xlf))
         import_sample_info(self.saf_num, self.bt)
         self.xrun = CustomizedRunEngine(self.bt)
+        # link mds
+        self.xrun.subscribe(xpd_configuration['db'].mds.insert, 'all')
```
CJ-Wright: I think you can just do `db.insert` now?

.travis.yml
chiahaoliu: It seems ``bluesky`` includes both version but from the [tag](https://github.com/NSLS-II/lightsource2-recipes/blob/master/scripts/nsls2-tag-build.sh#L52) it seems py35 is still the major platform now. I would prefer to test both, since ``travis`` is free :P

chiahaoliu:
yes, it will definitely great to put in under git control [issue here](https://github.com/xpdAcq/Roll-Out/issues/4). I would suggest do it after we test at bealine on Monday

PR #414, opened 2017-08-29
DOC: with doctr

CJ-Wright:


CJ-Wright:
@chiahaoliu ready to go!

CJ-Wright:
Note that these (once I enable the web page will send the files to https://xpdacq.github.io/xpdAcq/ rather than the current web page. We may want to deal with this at some point. I'm not exactly certain how we want to break up the project wide docs and the repo level docs. The DAMA docs may help with this.

PR #437, opened 2017-10-02
ENH: fix rever and add version number to docs

CJ-Wright:


setup.py
```diff
@@ -2,21 +2,9 @@
...
-            ver = ast.literal_eval(vstr)
-            break
-    return ver
-
-
 setup(
     name='xpdacq',
-    version=find_version('xpdacq/__init__.py'),
+    version='0.6.0',
     packages=find_packages(),
```
chiahaoliu: Are you reverting?

CJ-Wright: No, what is the current version number? (It will get bumped when we do the next tag)

chiahaoliu: I see. nvm.

0.6.0 is the current version. I thought ``find_version`` will deal with the version automatically.

CJ-Wright: `find_version` is super buggy, @scopatz and I decided that it would be better if we just let rever do its thing on the `setup.py` as well as the `__init__.py`. This PR is designed to go in before the tag.

PR #457, opened 2017-10-30
FIX: pure getx3 parser

chiahaoliu:
Refactoring on the md parser. Use ``diffpy.getX3`` as final parser, this should deal with edge case of the composition information we had before.


chiahaoliu:
yup, just included.

PR #458, opened 2017-11-03
DEP: insert mask in xrun

### LINKED ISSUE #455: MTN: remove print statement about masks
We don't do masks in Acq anymore so we don't need to warn about masks/not masks

chiahaoliu:
1. Deprecate print statement regarding inject mask inside ``xrun``, since mask is built dynamically
2. Update acq code corresponding to the API changes from bluesky 1.0

closes #455

### DIFF
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -162,7 +162,7 @@ def ct(dets, exposure):
                         'sp_uid': str(uuid.uuid4()),
                         'sp_plan_name': 'ct'})
     plan = bp.count([area_det], md=_md)
-    plan = bp.subs_wrapper(plan, LiveTable([area_det]))
+    plan = bp.subs_wrapper(plan, LiveTable([]))
     yield from plan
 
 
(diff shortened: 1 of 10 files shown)


xpdacq/beamtime.py
```diff
@@ -461,7 +462,7 @@ def statTramp(dets, exposure, Tstart, Tstop, Tstep, sample_mapping, *,
 register_plan('Tramp', Tramp)
 register_plan('tseries', tseries)
 register_plan('Tlist', Tlist)
-register_plan('statTramp', statTramp)
+#register_plan('statTramp', statTramp)
```
CJ-Wright: We might want a function (?) that loads up a bunch of plans (eg, "I'm running the cryostat so I need to load up all the cryostat runs"). This doesn't need to be in this PR.

chiahaoliu: Yeah, that's good idea. I think it might make sense to put all those scripts in a centralized location or repo and give user the option to load/import more complicated function.

Create an issue at #462

PR #461, opened 2017-12-04
ENH: update rever.xsh

CJ-Wright:


CJ-Wright:
@sbillinge this includes the new release statement

CJ-Wright:
Note that at some point we should move this `rst` file to some central place since we're going to put it into all the releases (and it is formatting friendly).

PR #463, opened 2017-12-11
FIX: mpl error with py3.6

chiahaoliu:
try to fix weird error from mpl when building py3.6


.travis.yml
```diff
@@ -37,9 +37,9 @@ before_install:
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - conda create -n testenv nose python=$TRAVIS_PYTHON_VERSION scipy scikit-beam jsonschema pytest coverage pip ophyd pyepics mock xlrd pyfai flake8 codecov ipython -c conda-forge -c lightsource2-tag -c soft-matter
+  - conda create -n testenv nose python=$TRAVIS_PYTHON_VERSION scipy numpy scikit-beam pytest coverage pip xlrd pyfai flake8 codecov ipython -c conda-forge
```
CJ-Wright: :+1:

.travis.yml
```diff
@@ -43,6 +43,7 @@ install:
   - 'pip install https://github.com/NSLS-II/bluesky/zipball/master#egg=bluesky'
   - 'pip install https://github.com/xpdAcq/xpdAn/zipball/master#egg=xpdan'
   - 'pip install https://github.com/xpdAcq/xpdSim/zipball/master#egg=xpdsim'
+  - pip install ophyd
```
CJ-Wright: fyi the conda-forge package for ophyd should be up to date.

requirements.txt
```diff
@@ -0,0 +1,13 @@
+attrs
+databroker
+doct
+doctr
```
CJ-Wright: Install doesn't require doctr

requirements.txt
```diff
@@ -0,0 +1,13 @@
...
+jinja2
+lmfit
+mongoquery
+networkx
```
CJ-Wright: Do we use networkx somewhere?

requirements.txt
```diff
@@ -0,0 +1,13 @@
...
+nose
+ophyd
+pims
+pyepics
+zmq
```
CJ-Wright: I think you want [pyzmq](https://pypi.python.org/pypi/pyzmq) also is this a requirement for xpdAcq?

CJ-Wright:
I think most of those requirements are from NSLS-II stack stuff (they should get pulled in when we install the stack). They were in the CI because we were building the NSLS-II stuff from scratch.

chiahaoliu:
Bluesky has two files specifying dependencies, dev-requirements.txt and requirements.txt. Only requirements.txt is called while installing bluesky. However, that only includes minimum requirements and it caused CI to fail during importing low level functions (not exactly remember which one but I can check the Travis log)

In xpdAcq, we do touch bluesky in a deeper level so I feel the need to have full stack installed which might reduce some troubles when we finally move to conda.

I agree current way of specifying full dependency is a bit too brittle. I will try to go over again and see if there is a better way of doing it.

CJ-Wright:
Are some of the NSLS-II deps listed as optional? If they are missing deps then maybe someone should fix that. If they are is there a way to tell pip to install all the optional deps too?

chiahaoliu:
superseded by #464

PR #465, opened 2018-01-05
Beamline config verification

### LINKED ISSUE #438: ENH: make long term beamline config more proactive
In an attempt to capture more metadata about the beamline config we should prompt BLS to change anything if needed
- [x] Put long term beamline config into start_beamtime function
- [x] Capture BLS initials when running start_beamtime
- [x] Print current config to screen (using `pprint` please)
- [x] Ask BLS to look it over and confirm config is correct.
- [x] If not correct pop up editor with config, then save and print/prompt again
- [x] if 'yes' continue with beamtime setup and save validated by information into config
- [ ] do one final load of config (with the validation info)

eaculb:
Closes #438 
Closes https://github.com/xpdAcq/mission-control/issues/43


eaculb:
@CJ-Wright having trouble putting the example configuration file in the correct location

CJ-Wright:
I think this is the issue. In the `conftest.py` (which sets up fixtures (all the state expected for a function to run) for the tests) an example config is missing, see [here](https://travis-ci.org/xpdAcq/xpdAcq/jobs/325564697#L1014). I think if you modify the `conftest.py` somewhere before [here](https://github.com/xpdAcq/xpdAcq/blob/master/xpdacq/tests/conftest.py#L53) to copy a sample config file to where it needs to go it should fix at least some of the tests. You may also want to make certain that the folders are deleted after the yield in the `conftest.py` see docs [here](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code)

xpdacq/tests/conftest.py
```diff
@@ -49,7 +49,7 @@ def bt(home_dir):
     saf_num = 300000
     wavelength = xpd_wavelength
     experimenters = [('van der Banerjee', 'S0ham', 1),
-                     ('Terban ', ' Max', 2)]
+                     ('Terban ', ' Max', 2)]
```
CJ-Wright: Why extra space?

eaculb:
@CJ-Wright @chiahaoliu Since we purposefully bypassed some of the code that asks for user input in order to successfully run the test, is it ok that this coverage test "failed"?

CJ-Wright:
That's fine, the coverage is a bit more like guidelines.

xpdacq/beamtimeSetup.py
```diff
@@ -33,7 +33,7 @@
 
 
 def _start_beamtime(PI_last, saf_num, experimenters=[],
-                    wavelength=None):
+                    wavelength=None,test=False):
```
CJ-Wright: Can you please put a space between these two kwargs? (PEP8)

xpdacq/beamtimeSetup.py
```diff
@@ -61,6 +61,7 @@ def _start_beamtime(PI_last, saf_num, experimenters=[],
         src = os.path.join(DATA_DIR, 'Ni24.D')
         dst = os.path.join(glbl_dict['usrAnalysis_dir'], 'Ni24.D')
         shutil.copy(src, dst)
+        beamline_config = _load_beamline_config(glbl['blconfig_path'],test=test)
```
CJ-Wright: Same here.

xpdacq/xpdacq_conf.py
```diff
@@ -180,14 +184,47 @@ def configure_frame_acq_time(new_frame_acq_time):
     print("INFO: area detector has been configured to new "
           "exposure_time = {}s".format(new_frame_acq_time))
 
-def _load_beamline_config(beamline_config_fp):
+def _verify_within_test(beamline_config_fp,verif):
+    while verif != "y":
```
CJ-Wright: Is there a reason to not do this with an `if` statement?

xpdacq/xpdacq_conf.py
```diff
@@ -180,14 +184,47 @@ def configure_frame_acq_time(new_frame_acq_time):
...
+        editor = os.getenv('EDITOR')
+    if not test:
+        while verif != "y":
+            with open(beamline_config_fp, 'r') as f:
+                beamline_config = yaml.load(f)
+            pp.pprint(beamline_config)
+            verif = input("\nIs this configuration correct? y/n: ")
+            if verif == "n":
+                print('Edit, save, and close the configuration file.\n')
+                os.system(editor + ' ' + beamline_config_fp)
```
CJ-Wright: you might consider using a subprocess call https://docs.python.org/3/library/subprocess.html

xpdacq/beamtimeSetup.py
eaculb: :+1:

xpdacq/beamtimeSetup.py
eaculb: :+1:

xpdacq/xpdacq_conf.py
eaculb: I just tried to leave it as close to the non-test syntax as possible, but can change it if you prefer.

CJ-Wright: No that's a fair point, its fine as is.

xpdacq/xpdacq_conf.py
eaculb: done

news/news.rst
```diff
@@ -1,13 +1,10 @@
...
-* Added news to the repo so we can changelog better
+* Verification step: beamline scientist must verify long-term beamline config when a new beamtime is started.
 
 **Changed:** None
 
-**Deprecated:** 
-
-* Remove static mask injection. Mask is now handled by the analysis
-  pipeline live.
+**Deprecated:** None
```
chiahaoliu: The news about deprecation of mask injection should be kept

xpdacq/xpdacq_conf.py
```diff
@@ -180,14 +185,47 @@ def configure_frame_acq_time(new_frame_acq_time):
...
-        beamline_config = yaml.load(f)
-
+    pp = pprint.PrettyPrinter()
+    os_type = platform.system()
+    if os_type == 'Windows':
+        editor = 'notepad'
+    else:
+        editor = os.getenv('EDITOR')
+    if not test:
+        while verif != "y":
```
chiahaoliu: It might not be a bad idea to accept ``Y`` in addition to ``y``(or even ``yes`` and ``Yes`` if you want to be flexible) since sometimes user will really type what they want, but not what we listed.

xpdacq/xpdacq_conf.py
```diff
@@ -180,14 +185,47 @@ def configure_frame_acq_time(new_frame_acq_time):
...
+            with open(beamline_config_fp, 'r') as f:
+                beamline_config = yaml.load(f)
+            pp.pprint(beamline_config)
+            verif = input("\nIs this configuration correct? y/n: ")
+            if verif == "n":
```
chiahaoliu: maybe we can handle this as  "if the answer is not ``y`` or ``Y``"?

xpdacq/xpdacq_conf.py
eaculb: Sure.

xpdacq/xpdacq_conf.py
eaculb: Do we not want to re-prompt if we don't get a clear yes or no?

news/news.rst
```diff
@@ -1,10 +1,14 @@
-**Added:** 
+**Added:**
```
CJ-Wright: This file shouldn't be named news, please go with something more descriptive for this branch

xpdacq/xpdacq_conf.py
chiahaoliu: maybe re-prompt if we don't receive an exactly yes?

xpdacq/xpdacq_conf.py
CJ-Wright: you might want to cast everything to lower/upper before checking it.

PR #467, opened 2018-01-08
ENH: customized message when xrun is interrupted

### LINKED ISSUE #466: FIX: correct syntax when xrun is interruputed
We should show correct syntax to the user when ``xrun`` is interrupted.

Currently, the default message from ``bluesky`` instructs the user to do ``RE.<method>`` but it should be ``xrun.<method>``.

chiahaoliu:
Replace ``RE`` with ``xrun`` in the printed message for correct syntax

closes #466


chiahaoliu:
@CJ-Wright thanks!

PR #468, opened 2018-01-09
MNT: print statement to run_calib.

### LINKED ISSUE #456: ENH: print in run_calibration
Put a print statement in `run_calibration` that states that users should look at the analysis terminal to click on rings and do other stuff.

chiahaoliu:
Instruct the user to look for analysis terminal and finish the interactive calibration process.

closes #456

### DIFF
--- a/xpdacq/calib.py
+++ b/xpdacq/calib.py
@@ -147,6 +147,14 @@ def run_calibration(exposure=5, dark_sub_bool=True,
                                     detector=detector,
                                     calibrant=calibrant
                                     )
+    print("INFO: Please navigate to the analysis terminal to complete "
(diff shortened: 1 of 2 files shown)


xpdacq/calib.py
```diff
@@ -147,6 +147,14 @@ def run_calibration(exposure=5, dark_sub_bool=True,
                                     detector=detector,
                                     calibrant=calibrant
                                     )
+    print("INFO: Please navigate to the analysis terminal to complete "
```
CJ-Wright: We should put a post it note on the analysis computer, saying "analysis computer" :smiling_imp:

PR #473, opened 2018-01-10
CLN: end_beamtime

### LINKED ISSUE #399: ENH: improve end_beamtime

### LINKED ISSUE #472: FIX: end beamtime
In light of the discussion at [mission-control](https://github.com/xpdAcq/mission-control/issues/42), we would like to fix the behavior of ``end_beamtime`` when there are files open and used by the process.

chiahaoliu:
This is another attempt to clean ``_end_beamtime``.

The clean introduced by this PR includes:

1. Use ``rsync`` while archiving ``xpdUser`` so that user can see the progress (``rsync`` lists files have been transferred.

2. More sophisticated logic when flushing ``xpdUser`` directory. Now the function will tell the user to close files used by the current process, instead of throwing an error and failing the process.

3. Some cleaning in the details. Program will remove remote archive if user doesn't confirm to flush local directory so that we could potentially avoid having multiple copies at the remote location. 

closes #472
closes #399

### DIFF
--- a/news/news.rst
+++ b/news/news.rst
@@ -21,4 +21,19 @@
   to tell the user to finish the interactive calibration process in the
   analysis terminal.
 
+* Fix ``_end_beamtime``. Details about the fixes are:
+
+  * Use rsync while archiving ``xpdUser`` so that user can see 
+    the progress. (rsync lists files have been transferred)
+
+  * More sophisticated logic when flushing xpdUser directory. 
+    Now the function will tell the user to close files used by 
+    the current process, instead of throwing an error and failing 
+    the process.
+
+  * Some cleaning in the logic. Program will remove the remote 
+    archive if user doesn't confirm to flush the local directory 
+    so that we could potentially avoid having multiple copies at 
+    the remote location.
+
 **Security:** None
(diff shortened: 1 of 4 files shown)


news/news.rst
```diff
@@ -11,6 +11,21 @@
 
 **Removed:** None
 
-**Fixed:** None
+**Fixed:**
+
+* Fix ``_end_beamtime``. Details about the fixes are:
```
CJ-Wright: Awesome!

PR #476, opened 2018-01-16
011618 xpd rollout

chiahaoliu:
Tested and verified locally at XPD.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -231,8 +231,8 @@ def _auto_load_calibration_file(in_scan=True):
 def _inject_filter_positions(msg):
     if msg.command == 'open_run':
         filter_bank = xpd_configuration['filter_bank']
-        filters = ['flt1', 'flt2', 'flt3', 'flt4']
-        print("INFO: Current filter status")       
+        filters = fb.read_attrs
+        print("INFO: Current filter status")
         for el in filters:
             print("INFO: {} : {}".format(el, getattr(filter_bank, el).value))
         msg.kwargs['filter_positions'] = {fltr: getattr(filter_bank, fltr).value for fltr in filters}
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -232,6 +232,9 @@ def _inject_filter_positions(msg):
     if msg.command == 'open_run':
         filter_bank = xpd_configuration['filter_bank']
         filters = ['flt1', 'flt2', 'flt3', 'flt4']
+        print("INFO: Current filter status")       
+        for el in filters:
+            print("INFO: {} : {}".format(el, getattr(filter_bank, el).value))
         msg.kwargs['filter_positions'] = {fltr: getattr(filter_bank, fltr).value for fltr in filters}
     return msg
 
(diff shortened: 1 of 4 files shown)


xpdacq/xpdacq.py
```diff
@@ -232,6 +232,9 @@ def _inject_filter_positions(msg):
     if msg.command == 'open_run':
         filter_bank = xpd_configuration['filter_bank']
         filters = ['flt1', 'flt2', 'flt3', 'flt4']
+        print("INFO: Current filter status")       
+        for el in filters:
+            print("INFO: {} : {}".format(el, getattr(filter_bank, el).value))
```
sbillinge: this is good.  Are we also saving this info by default as run metadata?  It may be in another PR, but I think that is behavior that we want.  Not just a print statement.

xpdacq/xpdacq.py
```diff
@@ -232,6 +232,9 @@ def _inject_filter_positions(msg):
     if msg.command == 'open_run':
         filter_bank = xpd_configuration['filter_bank']
         filters = ['flt1', 'flt2', 'flt3', 'flt4']
```
sbillinge: is it a bit dangerous hard-coding like this?  What happens if they add a fifth filter at the beamline?  This is ok for now, but somehow I feel it is better if there is a beamline config and this gets info from there.  Then, if the beamline config changes, all the code functionality will still work.  This will be rather relevant when we roll this out to PDF-the beamline.  There are probably other places in the code like this.  It may be a somewhat large refactor.  Only worth it if this code is going to be adopted at other beamlines.

chiahaoliu: Yes, I agree hard-code attribute names is bit dangerous. I changed the way of accessing attributes with ``read_attrs`` method from ``ophyd`` so that the code would work if the number of filters changes in the future.  @eaculb please let me know if you like this change.

Regarding the dependency of the code, I guess there might be two levels:

 1) pure software level (like file path etc): This can be properly handled if we migrate to other beamline. (It is currently managed by ``glbl`` but we can move to beamline config if needed)

2) instrument-software interface: This depends on which devices are used at the beamline and might be beamline specific. (for example if the filter in PDF beamline is not an ``ophyd.Device`` then subtle changes in syntax is needed)

I don't have good idea on how to properly separate those decencies yet but I will start an issue on mission-control so that we can start to discuss about it.

xpdacq/xpdacq.py
chiahaoliu: Yes, we inject the filter position into ``run_start`` metadata as well. Print statement is for experimenter.

xpdacq/xpdacq.py
eaculb: @chiahaoliu Yes, thank you, I agree that's much better.

xpdacq/xpdacq.py
```diff
@@ -231,7 +231,10 @@ def _auto_load_calibration_file(in_scan=True):
 def _inject_filter_positions(msg):
     if msg.command == 'open_run':
         filter_bank = xpd_configuration['filter_bank']
-        filters = ['flt1', 'flt2', 'flt3', 'flt4']
+        filters = filter_bank.read_attrs
```
CJ-Wright: :+1:

PR #480, opened 2018-01-20
MNT: bump up package version in __init__.py (used by doc)

chiahaoliu:


CJ-Wright:
Hmm that should have been in the rever bump, I guess I should look at that again.

PR #505, opened 2018-04-26
DOC: add docs for folder tag list

CJ-Wright:


CJ-Wright:
@sbillinge @chiahaoliu can I get some review?

docs/templates/usb_Running.rst
```diff
@@ -718,6 +718,10 @@ like ``folder_tag``. When ``folder_tag`` is added
 (eg ``xrun(0, 8, folder_tag='after_saxs')`` then the ``after_saxs`` will be
 added to the file path where the data is saved
 (eg ``<.../tiff_base/sample_name/after_saxs/mask/file_name>``).
+You can also define multiple folder tags by supplying ``folder_tag_list``
+(eg ``xrun(0, 8, folder_tag_list=['status', 'sample_name', bt_piLast'], status='after_saxs')``
+adding the values of the metadata for those keys into the file path
+(eg ``<.../tiff_base/after/saxs/sample_name/bt_piLast/mask/file_name>``).
```
chiahaoliu: is ``folder_tag_list`` a list of keys which defines the mapping of actual directory structure? Like ``sample_name`` is ``Ni`` and ``bt_pilast`` is ``Billinge`` for a certain scan and those fields are different for other scans. 

If this is the case, can we add ``<>`` around the keys which will be parsed later? and maybe we need some short explanation for this logic.

Also, in this example, shouldn't the folder name be:
``<.../tiff_base/after_saxs/...>``

CJ-Wright: Yes.

What do you mean "`<>` around the keys..."?

Yes, that is a typo.

CJ-Wright: Ah ok, I've updated the example, does it do a better job explaining now?

PR #513, opened 2018-06-04
FIX: robot issues at XPD

CJ-Wright:


CJ-Wright:
This needs tests I think (since we didn't catch them last time).

CJ-Wright:
@chiahaoliu I think this needs to go in (I'll fix the conflicts)

xpdacq/beamtime.py
```diff
@@ -601,6 +601,11 @@ def md(self):
         """ metadata of current object """
         return dict(self)
 
+    @property
+    def robot_all(self):
+        """All samples in the robot magazine"""
```
chiahaoliu: I guess this property is to show all samples in the magazine. Could you change to something like `show_all_sample_in_magzine` or other better name?

CJ-Wright: Sure

CJ-Wright: I removed the show since the primary intent of this was to be used in `xrun` as a place holder for all the samples (eg, run a robot scan on all the samples)

chiahaoliu: sure. makes sense.

CJ-Wright:
@chiahaoliu this is failing due to doc building, can we merge?

PR #519, opened 2018-06-15
Multi sample

CJ-Wright:
This supports multi-sample non-robot scans. This could work where there are multiple samples in a capillary, a multi-sample holder, or multi region thin films.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -622,9 +622,6 @@ class CustomizedRunEngine(RunEngine):
                                'global metadata')
         if robot:
             metadata_kw.update(robot=True)
-        if not robot and isinstance(sample, list):
-            raise RuntimeError('Multiple samples is not supported without'
-                               'the robot')
         # The CustomizedRunEngine knows about a Beamtime object, and it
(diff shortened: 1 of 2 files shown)


xpdacq/xpdacq.py
```diff
@@ -622,9 +622,6 @@ def __call__(self, sample, plan, subs=None, *,
                                'global metadata')
         if robot:
             metadata_kw.update(robot=True)
-        if not robot and isinstance(sample, list):
```
sbillinge: why was this even in?  Did we put it there for some expectation management or sthg?

CJ-Wright: Yes I think we weren't going to get the chance to test this so we didn't want people doing it by accident.

sbillinge: gotcha

sbillinge: @CJ-Wright do you want this pulled in?

CJ-Wright: yes please!

CJ-Wright:
Thank you!

PR #521, opened 2018-08-03
Add plan for tuning filters

CJ-Wright:


CJ-Wright:
This is very much a prototype, I'd like to test it before merging this but I thought I'd put it up for review so we could see if there were major issues.

CJ-Wright:
One potential issue here is that we are tuning the dark subtracted image. However, the detector burn might not care about the dark subtraction and only be based on the light image.

xpdacq/plans.py
```diff
@@ -0,0 +1,126 @@
...
+        computed_exposure = upper_threshold / (sp * attenuations[fci])
+        print("Extrapolated exposure: {}".format(computed_exposure))
+        exposure = min(5, computed_exposure)
+        print("Using exposure: {}".format(computed_exposure))
+        expected_counts = sp * attenuations[fci] * exposure
+        print("Expected counts: {}".format(expected_counts))
+        yield from bps.abs_set(
+            xpd_configuration["area_det"].cam.acquire_time, exposure
+        )
+
```
CJ-Wright: Run another round to double check.

sbillinge: reviewed on train together with CJ.  Looks promising, needs testing at the beamline.

We need an implementation/release plan when it is tested....I would like it to be largely hidden functionality and not change (enormously) the UI but this needs to be discussed. (we will need an override kwarg)

CJ-Wright: Also waiting a lightweight databroker for loading the light/dark data live.

CJ-Wright: xref: https://github.com/NSLS-II/bluesky/issues/1079

CJ-Wright:
@sbillinge @chiahaoliu ready for another round!

CJ-Wright:
This fails because the filler callback expects to get a start doc, but it doesn't because it is subscribed after the start is issued.

PR #526, opened 2018-09-24
REF: end_beamtime

chiahaoliu:
Refactor `end_beamtime` process to avoid the bottleneck at file transfer within NSLS-II network.

Detailed discussion and use case is located at

https://github.com/xpdAcq/mission-control/issues/135

cc: @dooryhee @sghose @adelezh

### DIFF
--- a/xpdacq/beamtimeSetup.py
+++ b/xpdacq/beamtimeSetup.py
@@ -244,11 +244,11 @@ def _end_beamtime(base_dir=None, archive_dir=None, bto=None, usr_confirm="y"):
     # load bt info
     archive_name = _load_bt_info(bto, _required_info)
     # archive file
-    archive_full_name = _tar_user_data(archive_name)
+    archive_full_name, local_archive_name = _tar_user_data(archive_name)
     # confirm archive
     _confirm_archive(archive_full_name)
     # flush
-    _delete_home_dir_tree()
+    _delete_archive_dir_tree(local_archive_name)
     # delete bt
     del ips.ns_table["user_global"]["bt"]
 
@@ -280,34 +280,33 @@ def _load_bt_info(bt_obj, required_fields):
 def _tar_user_data(archive_name, root_dir=None, archive_format="tar"):
     """ Create a remote tarball of all user folders under xpdUser directory
     """
-    archive_full_name = os.path.join(glbl_dict["archive_dir"], archive_name)
+    remote_archive_name = os.path.join(glbl_dict["archive_dir"],
+                                       archive_name)
     if root_dir is None:
         root_dir = glbl_dict["base"]
     try:
         os.chdir(root_dir)
-        print(
-            "INFO: Archiving your data now. That may take several"
-            " minutes. Please be patient :)"
-        )
+        # rename xpdUser to <xpdUser_archivefallname>
+        src = glbl_dict['home']
+        local_archive_name = '_'.join([src, archive_name])
+        local_archive_name = shutil.move(src, local_archive_name)
+        # make a fresh xpdUser dir. raise error if renaming wasn't successful
+        os.makedirs(src)
+        assert not os.listdir(src)  # assert empty
+        print("INFO: Archiving your data now. That may take several"
+              " minutes. Please be patient :)")
         # remove dir structure would be:
         # <remote>/<PI_last+uid>/xpdUser/....
-        os.makedirs(archive_full_name, exist_ok=True)
-        rv = subprocess.run(
-            [
-                "rsync",
-                "-av",
-                "--timeout=60",
-                "--no-owner",
-                "--no-group",
-                "--exclude=.*",  # exclude all hidden files
-                glbl_dict["home"],
-                archive_full_name,
-            ],
-            check=True,
-        )
+        os.makedirs(remote_archive_name, exist_ok=True)
+        rv = subprocess.run(['rsync', '-av', '--timeout=60',
+                             '--no-owner', '--no-group',
+                             '--exclude=.*',  # exclude all hidden files
+                             local_archive_name, remote_archive_name],
+                             check=True,)
     finally:
-        os.chdir(glbl_dict["home"])
-    return archive_full_name
+        # move back to fresh xpdUser
+        os.chdir(src)
+    return remote_archive_name, local_archive_name
 
 
 def _load_bt(bt_yaml_path):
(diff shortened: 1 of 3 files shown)


CJ-Wright:
LGTM, needs news.

chiahaoliu:
yup, just did.

xpdacq/beamtimeSetup.py
```diff
@@ -220,16 +220,15 @@ def _end_beamtime(base_dir=None, archive_dir=None, bto=None, usr_confirm='y'):
...
     # load bt info
     archive_name = _load_bt_info(bto, _required_info)
     # archive file
-    archive_full_name = _tar_user_data(archive_name)
+    archive_full_name, local_archive_name = _tar_user_data(archive_name)
     # confirm archive
     _confirm_archive(archive_full_name)
     # flush
-    _delete_home_dir_tree()
+    _delete_archive_dir_tree(local_archive_name)
```
sbillinge: this function name is confusing.  It looks as if you are deleting the archive.

maybe change it to ``_delete_local_archive``?

xpdacq/beamtimeSetup.py
```diff
@@ -259,25 +258,33 @@ def _load_bt_info(bt_obj, required_fields):
...
+    remote_archive_name = os.path.join(glbl_dict['archive_dir'],
+                                       archive_name)
     if root_dir is None:
         root_dir = glbl_dict['base']
     try:
         os.chdir(root_dir)
+        # rename xpdUser to <xpdUser_archivefallname>
+        src = glbl_dict['home']
+        local_archive_name = '_'.join([src, archive_name])
+        local_archive_name = shutil.move(src, local_archive_name)
```
sbillinge: I don't understand why this is done recursively.  why is it ``local_archive_name = shutil.move(src, local_archive_name)``.  What does ``shutil.move`` return?  Also I am not sure what is in ``glbl_dict['home']``. I suppose that is ``xpdUser``?  Or is it `~`. If the former, the line above seems to suggest that you are putting the archive into ``xpdUser`` which doesn't seem to be right.

xpdacq/beamtimeSetup.py
```diff
@@ -315,9 +322,7 @@ def _confirm_archive(archive_f_name):
         sys.exit(_graceful_exit("xpdUser directory delete operation "
                                 "cancelled at Users request."))
 
-def _delete_home_dir_tree():
-    os.chdir(glbl_dict['base'])  # move out from xpdUser before deletion
-    dir_to_flush = glbl_dict['home']
+def _delete_archive_dir_tree(dir_to_flush):
```
sbillinge: see above

xpdacq/tests/test_beamtimeSetup.py
```diff
@@ -10,7 +10,7 @@
 from xpdacq.beamtime import Beamtime, ScanPlan
 import xpdacq.beamtimeSetup as bts
 from xpdacq.beamtimeSetup import (_start_beamtime, _end_beamtime,
-                                  _delete_home_dir_tree, _make_clean_env,
+                                  _delete_archive_dir_tree, _make_clean_env,
```
sbillinge: see above

xpdacq/beamtimeSetup.py
chiahaoliu: yeah, agree. I will update the name.

xpdacq/beamtimeSetup.py
chiahaoliu: `local_archive_name` on the left side (new) is the return value from the `shutil.move` function which tells us where is the file actually moved to, with respect to the file system. 

I overwrite the `local_archive_name` (old) on the right side with the new one just in case if `xpdUser` is placed at some weird location in the future (and possible symlic back), `shutil` can still help us figure out the absolute path on the local file system.

Yes, `glbl['home']` is `<nfs_mount_path>/xpdUser`. I am renaming `xpdUser`, loaded with user data to `xpdUser_<user beamtime_indentifier>` so that the next beamtime won't be blocked by the archiving process.

sbillinge:
ok, I see

xpdacq/beamtimeSetup.py
sbillinge: OK, this makes sense.  But please check what the archive will be called.  You are wanting ``xpdUser_<user beamtime id>`` but based on what you have written above, you may end up with ``~\this_and_that\xpdUser_<user beamtime id>``

chiahaoliu:
Yeah, that's a good point. Let me explain more to see if it makes sense.

We need full path (`~/this_and_that/xpdUser_<PI_last>_<saf_num>_<bt_uid>`) after renaming since it's used in the step of transfering local archive to remote location via `rsync`. I also make sure the parent path of renamed dir matches with orignal parent path of `xpdUser`. Archive will be placed in following directory `/archive_location/YYYY/<PI_last>_<saf_num>_<bt_uid>/`.

adelezh:
Any updates please? can we include this issue in 2019 cycle 1 task?

CJ-Wright:
I'm going to try to patch this up (get rid of the merge conflicts and get travis to pass) and merge it. It is not the most optional solution, as @DanOlds says it would be nice to have this happening continuously, but this is a complete solution which seems to work. If we have a continuously saving solution then it should be possible to merge these two things (eg save to the archive dir the data, remove the `tiff_base` folder from xpdUser (once they are done with it) and then move the other folders over, finally compressing it).

sbillinge:
Thanks @CJ-Wright .  @DanOlds, I think we need a concrete plan for what you
want saved where for the real-time saving (which is  a great idea btw), and
we can code it up.  Could we assign to you the task of capturing that in
all its gory detail, and it should be straightforwrad to implement, even
mid-cycle, at least for testing.

 I guess it imeans getting input from all the beamline staff about what
they want saved where.  Things are easier and more robust now we have the
0MQ server as we can dump things all over the place by setting up different
write servers.

The original philosophy for end-beamtime was that
(a) we would guarantee a clean environment to each new user (they also
wouldn't be nervrous that subsequent users would see their stuff, at least
easily).
(b) When they leave the beamline we could guarantee that everything they
did at the beamline woudl be archived and could be recovered.

I think that this philosophy has been a success (no user complaints about
this) apart from hassles now emerging of slow change beamtimes.  We have to
fix this, but we should be sure we don't throw out the baby with the
bathwater.....

I am not sure whether we still need to offer (b) as a guarantee, but I
would prefer that any solution has safeguards that data are not
inadvertently lost.  This is less likely now as we have pipelines that can
be rerun post-facto.  We can design a new end-beamtime sequuence but it
would be nice to be really clear about what the goals/objectives are.

S

DanOlds:
My thought is that we should not be backing up at the end of a beamtime,
but rather the whole time.  It seems to me a natural sokution for that if
we can now write simultaneous copies of files in real time.  This way, end
beamtime really only needs to clean up for the next user.

I am still working on resolving our storage solution for this cycle.  The
default, which is a very poor default in my opinion, would be the same
location as last cycle.  I will continue to push on our IT group.

Dan

CJ-Wright:
This is pending an xpdAn tag

PR #528, opened 2018-09-27
use sa_uid not uid

CJ-Wright:


CJ-Wright:
These were fixed by #513

PR #532, opened 2018-11-26
add shutter sleep

CJ-Wright:
Attn: @DanOlds 
@chiahaoliu The idea is to add a sleep to the shutter stuff so the in-situ scans work.


chiahaoliu:
Looks good to me! Thanks.

PR #544, opened 2019-03-20
Force archive directory creation

CJ-Wright:


CJ-Wright:
@chiahaoliu I think this fixes the issue out at PDF where the archive dir didn't exist so maybe rsync balked?

chiahaoliu:
@CJ-Wright I don't think this is the case tho. We have the same thing at [line 300](https://github.com/xpdAcq/xpdAcq/blob/master/xpdacq/beamtimeSetup.py#L300)

PR #547, opened 2019-03-29
put close stub in take_dark

CJ-Wright:


chiahaoliu:
Could we also make it `_close_shutter_stub` and `_open_shutter_stub` not priviate?

st3107:
New dark method will be implemented. We will not refactor the current method unless necessary since it is going to be deprecated. This PR is closed.

PR #553, opened 2019-09-03
FIX: pyyaml.load

### LINKED ISSUE #543: PyYAML 5 breakage
It seems that PyYAML 5 causes breakage for us?

chiahaoliu:
closes #543 

use loader suggested by pyyaml team

### DIFF
--- a/docs/templates/db_export.rst
+++ b/docs/templates/db_export.rst
@@ -60,7 +60,7 @@ The hard way
         """
     with open(os.path.join(p, 'raw.yml'), 'w') as f:
         f.write(portable_template)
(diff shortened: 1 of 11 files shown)


chiahaoliu:
@CJ-Wright ready for review. 

Note: as scary as it might seem, `unsafe_load` simply means allowing loading an python object that is not in built-in data structure. In our case, we have d-spacing as a`numpy.array`. This behavior was allowed by default in pyyaml <= 3.13. This approach is also suggested by pyyaml dev team: https://github.com/yaml/pyyaml/issues/286

CJ-Wright:
LGTM

PR #559, opened 2020-08-30
BLD: add dependencies to setup.py

tacaswell:
I am not sure that these are correct, please check and adjust before merging.


st3107:
Thank you @tacaswell . I will check that. It seems that db is missing from the xpdsim and that's wht the tests failed. I will first look at xpdsim and then turn back to this PR.

st3107 (approved):

PR #560, opened 2020-08-30
ENH/MNT: pull out stateless function to configure exposure time

tacaswell:
I think I shimmed everything back to be the same as it was.

The key idea here is to separate the "this is how a detector is configured" from "these is the global state that xpdacq manages".  By separating them we can independently change either the plan used or how we manage the global state which is going to become increasingly important as e move to a remote-first model.

### DIFF
--- a/xpdacq/beamtime.py
+++ b/xpdacq/beamtime.py
@@ -80,18 +80,27 @@ def _summarize(plan):
     return "\n".join(output)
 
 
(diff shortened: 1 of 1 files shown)


st3107:
Hi @tacaswell I wonder if this PR is ready for review.

st3107:
Continue in PR #568

PR #568, opened 2020-09-16
ENH/MNT: pull out stateless function to configure exposure time

st3107:
Continue PR #560.


st3107:
@tacaswell Thank you. I moved your PR here and did some refraction. You pointed out an important issue in current xpdacq. It  tends to abuse the global variable xpd_configuration, and glbl.

tacaswell:
:tada:

PR #580, opened 2021-01-09
FIX imports bugs in xpdacq.calib and FIX sample metadata bugs in the xpdacq.xpdacq

st3107:


gwbischof:
Thanks for the fix!

sbillinge (commented):
Thanks so much for this @st3107 

fyi, we seem to have messed up the TEMPLATE.rst.  We may need another PR to fix it?  Also, to make the news compile more nicely it is better to do it this way...
```
**Added:**
 * A tool to inject metadata using the preprocessors.
 * Add dependency on the pyopenxl
```
(note the whitespace at the beginning of the line with the bullet).  

Also, put back the None when there is no news item in a category.

PR #589, opened 2023-03-23
Fix/donot stomp on user calibration md

tacaswell:
This caused us great pain at PDF today.

### DIFF
--- a/xpdacq/xpdacq.py
+++ b/xpdacq/xpdacq.py
@@ -633,12 +633,10 @@ def _inject_calibration_md(msg):
     if msg.command == "open_run":
         exp_hash_uid = glbl.get("exp_hash_uid")
         # inject client uid to all runs
-        msg.kwargs.update({"detector_calibration_client_uid": exp_hash_uid})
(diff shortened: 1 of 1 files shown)


sbillinge:
do you need this merged?  It is failing tests though

tacaswell:
We monkey patched our way around the problem at PDF.  The test failures looks like insufficiently pinned databroker / tiled interactions....

sbillinge:
@tacaswell I will need a bit more help to fix this properly.  it is giving an import error in the CI.
`ImportError: cannot import name 'from_tree' from 'tiled.client' (/usr/share/miniconda/envs/test/lib/python3.10/site-packages/tiled/client/__init__.py)`

I guess that tiled  has shifted and we need to update our end to follow?

st3107:
@sbillinge I think it is OK to merge and let workflow fail. The temp in the new databroker may not work. We need to wait for some development for the stable temp database feature. The fix from Tom is definitely going to work so it is fine to merge and release the patch.

